PR #881, opened 2022-09-27
Inject providers to `window` in such a way they can't be reassigned

steveluscher:
Addresses #882.


steveluscher:
(should have maybe mentioned that I literally didn't test this) 🤞🏻

Issue #727, opened 2022-09-01
make ethereum token send gas settings configurable


tomlinton:
Gas settings and nonce should be configurable. They are all passed [from the client ](https://github.com/coral-xyz/backpack/blob/7dc1b2381e786eca2b671b40f20ba0db84d44fa2/packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx#L559-L563) so we just need a configuration screen. I think the existing drawer is too small so might need a full screen drawer for these settings.

PR #889, opened 2022-09-27
configurable ethereum transactions

tomlinton:
Implements Ethereum transaction configuration options wherever they are sent (injected provider approvals, send, xnft approvals). Not needed for swap transactions yet as no Ethereum support, but will be eventually.

Not an exact match for Figma. Will raise issues for the following:

- Find an API or write a formula to estimate transaction time based on historical gas prices and confirmation times.
- Add info icons and tooltips for each of the settings in the transaction window.
- More form validation/feedback for the settings.

Closes https://github.com/coral-xyz/backpack/issues/727


packages/app-extension/src/components/Unlocked/ApproveTransactionRequest.tsx
```diff
@@ -223,19 +222,7 @@ function SignAllTransactionsRequest({
...
-      onResolve={onResolve}
-      onReject={onReject}
-      loading={false}
-      transactionToSend={transactions}
-      from={walletPublicKey.toString()}
-      network={"Solana"}
-      networkFee={"-"}
-    />
-  );
+  return <></>;
```
tomlinton, line 225: TODO

PR #891, opened 2022-09-27
common component for header token amount display

tomlinton:
Ensures a consistent display among all these header balances that we had (e.g. we had ellipsis for long values in some places and not others). Also changed to JS for truncation because the font isn't monospace so space between the ellipsis and ticker can be inconsistent.


trucy:
@tomlinton Inter has OpenType [features](https://rsms.me/inter/#features) and a setting to display as monospace. That might help the display of these more consistent.

I think it's something like 

`font-feature-settings: ‘tnum’ on`

tomlinton:
Thanks @trucy! I can enable that. It does look pretty different though so more of a design question than needed for this I think.

PR #904, opened 2022-09-28
Catches transaction errors in ApproveTransactionRequest

### LINKED ISSUE #810: xnft: surface failed simulation errors in transaction approval
See https://discord.com/channels/985994296337498182/1011846320677462137/1019161472397488249

hkirat:
The PR catches errors in `ApproveTransactionRequest` that is causing unhandled rejections to remain uncaught. This prevents application developers from knowing if the underlying transaction failed.

I've now called the `onReject` callback in case the background requests fails which will cause `window.xnft.solana.send` throw, which the application developer can then handle according to their app.

Does it make sense to propagate the `error message` to the developer? Right now the `error message` that reaches them is `user rejected signature request` from [here](https://github.com/coral-xyz/backpack/blob/master/packages/app-extension/src/components/Unlocked/ApproveTransactionRequest.tsx#L124)

Closes #810

### DIFF
--- a/packages/app-extension/src/components/Unlocked/ApproveTransactionRequest.tsx
+++ b/packages/app-extension/src/components/Unlocked/ApproveTransactionRequest.tsx
@@ -312,12 +312,14 @@ function _SendTransactionRequest({
   // into this component because it can be modified by the user to set
   // transaction specific settings (i.e. Etheruem gas).
   //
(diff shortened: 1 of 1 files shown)


armaniferrante:
We should propagate the error message to the developer. We can do that in a future change.

PR #919, opened 2022-09-29
Support for Windows/Powershell builds.

skynetcap:
- Use `cross-env` devDependency for setting NODE_ENV, which supports windows. 
- Use node-native copy command instead of `cp` for cross-platform support.


skynetcap:
Updated PR to use `cross-env` devDependency, instead of relying on `set` functionality which resulted in a reported build error.

tomlinton:
Thanks for testing this @skynetcap! Similarly the [shx](https://www.npmjs.com/package/shx) package would be good for the cp command as it is probably a bit more readable.

skynetcap:
> Thanks for testing this @skynetcap! Similarly the [shx](https://www.npmjs.com/package/shx) package would be good for the cp command as it is probably a bit more readable.

Addressed by [529b871](https://github.com/coral-xyz/backpack/pull/919/commits/529b87160a0d8fecb07593b655d47234b38bcd6a)

armaniferrante:
Thanks again @skynetcap. Sorry about the delay getting it in (had to address some unrelated build issues while traveling).

Issue #882, opened 2022-09-27
Rewrite providers to be immutable to prevent certain classes of MITM attack


steveluscher:
#### Problem statement

Third party code with access to the same `window` sandbox as Backpack can override the Backpack providers. Threats include, but are not limited to, MITM attacks:

```typescript
const oldSendTransaction = window.xnft.solana.send.bind(window.xnft.solana);
window.xnft.solana.send = (transaction) => {
  // Maliciously modify transaction.
  const evilTransaction = getModifiedTransaction(transaction);
  // Forward it to the transaction sender.
  return oldSendTransaction(evilTransaction);
};
```

#### Proposed solution

- [x] #881 
- [ ] Rewrite providers using private members and deep freezing
    ```typescript
    class ProviderThing {
      #publicKey = null  // JavaScript private member
      constructor() {
        if(new.target === ProviderThing) {
          Object.freeze(this);  // Deep freeze everything but private members
        }
      }
      getPublicKey() {
        return this.#publicKey;
      }
      setPublicKey(value) {
        this.#publicKey = value;  // Still works, despite being frozen.
      }
    }
    ```

steveluscher:
I don't intend on sending a PR for the “rewrite providers using private members and deep freezing” step. If you, dear reader, would like to step up and perform that refactor, go for it!

hkirat:
Can take this up if no one else is already working on it.

PR #925, opened 2022-09-29
Rewrite providers using private members and add deep freezing

hkirat:
The PR makes all attributes `private` in the providers to avoid MITM attacks. More context in https://github.com/coral-xyz/backpack/issues/882

I've made all the attributes private and added `get`'s and `set`'s for attributes we want publicly available to keep the public API of all the providers the same.

So the publicly available attributes have the following gets and sets associated with them - 
```js
  get publicKey() {
    return this.#publicKey;
  }
  set publicKey(value) {
    this.#publicKey = value;
  }

```

This is slightly different from the suggested implementation in #882 - 
```js
  getPublicKey() {
    return this.#publicKey;
  }
  setPublicKey(value) {
    this.#publicKey = value; 
  }
```

Closes https://github.com/coral-xyz/backpack/issues/882


packages/provider-core/src/provider-ethereum-xnft.ts
```diff
@@ -124,17 +128,41 @@ export class ProviderEthereumXnftInjection extends EventEmitter {
...
+    this.#publicKey = publicKey;
+  }
+
+  public set connectionUrl(connectionUrl) {
+    this.#connectionUrl = connectionUrl;
+  }
+
+  public set provider(provider) {
+    this.#provider = provider;
+  }
```
steveluscher, line 167: Do we actually _want_ these setters, or was the ability to set these just an accident of them being public before this PR?

packages/provider-core/src/provider-ethereum.ts
```diff
@@ -350,10 +354,10 @@ export class ProviderEthereumInjection extends EventEmitter {
   async _handleNotificationDisconnected() {
     if (this.isConnected()) {
       // Reset public state
-      this.chainId = null;
-      this.publicKey = null;
+      this.#chainId = null;
+      this.#publicKey = null;
       // Reset private state
-      this._state = {
+      this.#state = {
```
steveluscher, line 360: Careful that this object is still mutable, if ever anything can get a hold of it. Consider `Object.freeze` here?

packages/provider-core/src/provider-ethereum.ts
```diff
@@ -386,19 +390,19 @@ export class ProviderEthereumInjection extends EventEmitter {
...
   }
 
   /**
    * Update local state and emit required event for connect.
    */
   protected async _handleConnect(chainId: string) {
-    if (!this._state.isConnected) {
-      this._state.isConnected = true;
+    if (!this.#state.isConnected) {
+      this.#state.isConnected = true;
```
steveluscher, line 405: If you decide to `Object.freeze` the state, then you could still accomplish mutation by:

```typescript
this.#state = {...this.#state, isConnected: true};
```

steveluscher (commented):
Love it! Not going to review this for real, but just left some thoughts, particularly about whether we want some of these things to be mutable from the outside at all.

packages/provider-core/src/provider-ethereum-xnft.ts
hkirat: Yeah this was mostly to provide backward compatibility for existing public attributes. 
I think it makes sense for these to generally be private.

steveluscher: Kill kill kill. I can't imagine it's desirable for randos to be able to patch these up at runtime from the outside. In general the interfaces should be thin thin thin.

packages/provider-core/src/provider-ethereum.ts
hkirat: Ah yeah great catch

hkirat: Done!

packages/provider-core/src/provider-ethereum-xnft.ts
hkirat: agreed agreed agreed!

packages/provider-core/src/provider-ethereum.ts
hkirat: Probably need to re-freeze it everytime I'm re-assigning it as well.

hkirat:
@steveluscher it feels like this can still be manipulated. 
The private variables have become inaccessible but a malicious user can still call functions on the class and modify the state. 
For eg: 
```js
window.xnft.solana._handleNotificationConnectionUrlUpdated({
  event: {
    data: {
       detail: {
         data: {
            url: "<incorrect-rpc-url>"
         }
       }
     }
  }
})
```

Not sure if this exact case could be malicious but we have a bunch of private handlers inside the providers which aren't truly private (can still be accessed from outside).

Does it make sense to convert these handlers to `truly private` functions as well?

Something like - 
```js
export class ProviderSolanaInjection  {
  #handleNotificationConnectionUrlUpdated = (event: Event) => {
    this.#connection = new BackgroundSolanaConnection(
      this.#connectionRequestManager,
      event.data.detail.data.url
    );
  }
}
```

hkirat:
hmm this might be slightly trickier than I initially anticipated. 
Turns out the providers extend `EventEmitter` which has `non private` fields it needs to update from time to time. 
Eg - [here](https://github.com/primus/eventemitter3/blob/master/index.js#L66) inside the `addListener` function we increase `_eventsCount` by `1` which will complain if the object is frozen.

Can think of two ways to get around this - 
1. Create a `CustomEventEmitter` that extends `EventEmitter` and fix up all the methods that will cause an issue if the object is frozen.
2. Create an inhouse implementation of EventEmitters

Is there another way I might be missing? Is this even needed or can someone think of a leaner solution?

steveluscher:
> Turns out the providers extend EventEmitter which has non private fields it needs to update from time to time.

Good god. I hate software.

The only hack I can think of off the top of my head is to proxy calls.

```typescript
class Foo implements EventEmitter3 {
  #emitterInstance: EventEmitter3;
  constructor() {
    this.#emitterInstance = new EE(/* ... */);
  }
  on(...args) {
    return this.#emitterInstance.on(...args);
  }
  off(...args) {
    return this.#emitterInstance.off(...args);
  }
  emit(...args) {
    return this.#emitterInstance.emit(...args);
  }
  /* ...etc... */
}
```

…but, like, gross.

hkirat:
I'm tempted to re-write a `PrivateEventEmitter` in TS (Option 2 above). I worry we'll have to keep it up to date though/maintain it as an independent project. 
Might just do it though unless there are any objections. cc @armaniferrante 
Your approach sounds pretty lean @steveluscher , might resort to that if `PrivateEventEmitter` turns out to be too much work/we don't want the burden of maintaining an internal fork of `EventEmitter`.

steveluscher:
Send 'em a PR and see what they think, maybe?

steveluscher:
Yeah! Private class features _are_ part of JavaScript, not Typescript. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_methods

All you have to do is to convince EE that it’s worth ratcheting up the runtime requirements.

armaniferrante:
@hkirat I'm inclined to go with @steveluscher's proxy suggestion, rather than option 2, as it sounds like less of a maintenance burden. Are there any notable advantages of option 2 over the proxy?

hkirat:
@steveluscher Ah okay that makes sense. 
@armaniferrante none that i can think of. I think that makes sense. Makes the change much more leaner as well.

packages/provider-core/src/provider-solana-xnft.ts
```diff
@@ -32,36 +31,40 @@ import {
...
+  #requestManager: RequestManager;
+  #connectionRequestManager: RequestManager;
 
-  public publicKey?: PublicKey;
-  public connection: Connection;
+  #publicKey?: PublicKey;
+  #connection: Connection;
 
-  constructor(requestManager: RequestManager) {
+  constructor(requestManager: RequestManager, freeze = true) {
```
hkirat, line 51: This `freeze` argument here feels slightly ugly. I introduced it because we first initialize `ProviderSolanaXnftInjection` in `provider-injection/src/index.ts` [here](https://github.com/coral-xyz/backpack/pull/925/files#diff-f002589cac573b8ca88b56a861269558b1b3b8756c8c88c1f99a764df55641bbR43) and when we try to update it using `Object.defineProperties` it complains.
So I freeze it after the assignment [here](https://github.com/coral-xyz/backpack/pull/925/files#diff-f002589cac573b8ca88b56a861269558b1b3b8756c8c88c1f99a764df55641bbR51)

packages/provider-core/src/provider-ethereum.ts
```diff
@@ -115,46 +115,49 @@ export class ProviderEthereumInjection extends EventEmitter {
...
+    this.#setState({
       ...ProviderEthereumInjection._defaultState,
-    };
-
-    this.isBackpack = true;
-    this.chainId = null;
-    this.publicKey = null;
+    });
 
-    this._handleConnect = this._handleConnect.bind(this);
```
hkirat, line 145: Removed this block and changed the function signature of all of these handlers to be `arrow functions` to bind `this`

hkirat:
Just tested this on both eth and solana using the `provider-test` example. Both seem to be working as expected.

packages/provider-core/src/provider-solana-xnft.ts
armaniferrante: 2 cents: Might be a little cleaner to just have two different classes. `ProviderSolanaXnftInjection` And `UnfrozenProviderSolanaXnftInjection`, where the latter has a `freeze()` method returning the former.

hkirat: Do you mean something like this - 

```ts
class UnfrozenProviderSolanaXnftInjection
{
   constructor() {
       // do everything, but dont freeze
   }
   public freeze() {
      return new ProviderSolanaXnftInjection(this.requestManager)
   }
   // All the existing functions

}

class ProviderSolanaXnftInjection extends UnfrozenProviderSolanaXnftInjection {
     constructor(requestManager) {
        super(requestManager);
        Object.freeze(this)
    }
}
```
?

steveluscher: No need! The way that you deep freeze subclasses is this:

```typescript
class SubclassOfFoo extends Foo {
  constructor() {
    if (new.target === SubclassOfFoo) {
      Object.freeze(this);
    }
  }
}
```

steveluscher: Oh, I see what you're saying above. In my opinion, the argument shouldn't be `freeze`; the argument should be the other stuff you want to define.

```typescript
class ProviderSolanaXnftInjection {
  constructor(requestManager, additionalProperties) {
    super(requestManager);
    const additionalPropertyConfig = {};
    Object.keys(additionalProperties).forEach(prop => {
      additionalPropertyConfig[prop] = {value: additionalProperties[prop]};
    });
    Object.defineProperties(this, additionalPropertyConfig);
    if (new.target === ProviderSolanaXnftInjection) {
      Object.freeze(this);
    }
  }
}

new ProviderSolanaXnftInjection(requestManager, {
  ethereum: new ProviderEthereumXnftInjection(requestManager),
  solana: new ProviderSolanaXnftInjection(requestManager),
});
```

steveluscher:
Left feedback for an alternate approach to defining properties that doesn't depend on doing a `freeze: true`/`freeze: false` dance.

hkirat:
A high level meta question (maybe unrelated to the ticket) - Why does `ProviderSolanaXnftInjection`get to be the root of the `xnft` object? vs a `ProiverRootXnftInjection` or even `ProviderEthereumXnftInjection` ?

armaniferrante:
> A high level meta question (maybe unrelated to the ticket) - Why does `ProviderSolanaXnftInjection`get to be the root of the `xnft` object? vs a `ProiverRootXnftInjection` or even `ProviderEthereumXnftInjection` ?

No good reason other than it was implemented first. I agree with your point.

armaniferrante:
Thanks a bunch for this lift @hkirat.

PR #931, opened 2022-09-30
share toggle password visibility button with confirmation field

wentokay:
it feels like having a single button shared between both fields is a bit simpler, but maybe a bad idea?

https://user-images.githubusercontent.com/101902546/193170469-97c0832e-c98b-4ca8-b45d-7fac51c67ffc.mov


tomlinton:
cc @trucy

wentokay:
current implementation for reference 


https://user-images.githubusercontent.com/101902546/193172071-3934e876-21ca-44ee-8558-b62aabed776c.mov

trucy:
lgtm!

PR #953, opened 2022-10-01
improve onboarding flow

wentokay:
- uses memoryrouter to simplify managing state between onboarding pages
- includes inline back button on each page
- includes hamburger menu on every onboarding page
- fixes margin and padding issues
- moves 'I saved' checkbox between copy and next buttons on mnemonic page as per design
- moves password page to after mnemonic input for both creating and importing wallet
- wraps forms in <form /> tag so that they can be submitted with enter key (in most cases)


packages/app-extension/src/app/App.css
```diff
@@ -38,8 +38,10 @@ body {
   left: 0px;
   right: 0px;
   bottom: 0px;
-  overflow: scroll;
+  overflow: hidden;
```
wentokay, line 41: unsure if this causes problems in the wallet extension itself, i didn't notice anything immediate

ka636-cyber: Panda power

armaniferrante:
Rebased master.

PR #959, opened 2022-10-02
Export useTabContext

ghulamMustafaRaza:
Export useTabContext so that it can be used for navigation by other UIs.


tpompon:
Maybe we could integrate it in `useNavigation` hook, but now this update update would help to build better custom navigation paths

ghulamMustafaRaza:
@tpompon yes we can also there should be a possibility to use Nested navigations either stack or tab. so Navigation API need updates anyways.

PR #987, opened 2022-10-03
make waiting room accessible to onboarded users

callensm:


packages/app-extension/src/components/Unlocked/Settings/index.tsx
```diff
@@ -759,6 +765,12 @@ function SettingsList({ close }: { close: () => void }) {
       icon: (props: any) => <AccountCircleOutlined {...props} />,
       detailIcon: <PushDetail />,
     },
+    {
+      label: "Waiting Room",
```
armaniferrante, line 769: Maybe we should move this into the bottom section, next to Discord?

PR #989, opened 2022-10-03
When validating secret key throw error if has already been added

colinlgray:
### What this does:
This displays a new error message to the user when they import a private key that has already been added

### What code changed:
The useWalletPublicKeys hook was added to the ImportSecretKey component. The validateSecretKey function was changed to accept the existing keys as a parameter, and to throw an error if the key is invalid. Previously the validateSecretKey function would return false to indicate an invalid secret key but now it throws an error

### Additional notes
Please let me know what additional information I should provide, changes that need to be made to the code, or additional screens that need to have this check added to them

### DIFF
--- a/packages/app-extension/src/components/Unlocked/Settings/index.tsx
+++ b/packages/app-extension/src/components/Unlocked/Settings/index.tsx
@@ -943,13 +943,15 @@ export function ImportSecretKey({ blockchain }: { blockchain: Blockchain }) {
   }, [theme]);
 
   const onClick = async () => {
-    const secretKeyHex = validateSecretKey(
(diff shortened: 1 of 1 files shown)


armaniferrante:
What happens in the UI with this change?

colinlgray:
Thanks for that, as I was testing how the UI changed I realized it wasn't right and made another change to allow the error message to be displayed properly. Here is a screenshot of the new message along with an example of the old message. Also I will update the PR description above to include this change

Screen Shot 2022-10-03 at 7 21 59 PM
Screen Shot 2022-10-03 at 7 22 08 PM

PR #1004, opened 2022-10-04
use root pubkey for authentication flow

wentokay:
will unblock #1003 once merged

TODO

- [ ] fix `await bc.signMessage(encode(buffer), publicKey!);` so that it works even if importing keys from a different derivation path (related #1024)


tomlinton:
We decided not to do this for now!

Issue #992, opened 2022-10-03
Eth transaction errors are not caught


hkirat:
Slightly related to https://github.com/coral-xyz/backpack/pull/904

If I try `Sending an Ethereum Transaction` from an empty wallet, there is an uncaught exception. 
<img width="1720" alt="Screenshot 2022-10-04 at 1 58 23 AM" src="https://user-images.githubusercontent.com/8079861/193676810-8c04382e-cc2e-4b22-bd7b-adab960e2b20.png">
<img width="367" alt="Screenshot 2022-10-04 at 1 58 18 AM" src="https://user-images.githubusercontent.com/8079861/193676849-67b5eb60-8b28-49bb-ac76-01ea0350a222.png">

This is different from https://github.com/coral-xyz/backpack/pull/904 where the exception occurs after the user has clicked on the approve button.

In this particular case the exception occurs even before the user clicks on the `Approve` button.

Not sure what the right UX is here - 
1. Should we immediately close the `Approval Popup` and surface the exception to the developer?
2. Should we maybe show a warning to the user that this transaction might fail even if you click on `approve` ? 
3. Should we just catch and silently ignore this exception?

armaniferrante:
We probably want to immediately show the error on the same Approval Popup drawer. Putting on design.

PR #1029, opened 2022-10-06
Fixed eth error when `estimateGas` fails

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/992

Turns out this happens because the `populateTransaction` call [here](https://github.com/coral-xyz/backpack/blob/master/packages/recoil/src/hooks/useTransactionData.tsx#L90) fails in `ethers.js` [here](https://github.com/ethers-io/ethers.js/blob/01aea705ce60b1c42d2f465b162cb339a0e94392/packages/abstract-signer/src.ts/index.ts#L301). 

I've added a default `gas limit` which we anyways overwrite later on [here](https://github.com/coral-xyz/backpack/blob/master/packages/recoil/src/hooks/useTransactionData.tsx#L118).

### DIFF
--- a/packages/recoil/src/hooks/useTransactionData.tsx
+++ b/packages/recoil/src/hooks/useTransactionData.tsx
@@ -8,7 +8,6 @@ import { Blockchain, UI_RPC_METHOD_SOLANA_SIMULATE } from "@coral-xyz/common";
 import {
   useBackgroundClient,
   useBlockchainNativeTokens,
-  useBlockchainTokensSorted,
   useEthereumCtx,
   useEthereumPrice,
   useSolanaCtx,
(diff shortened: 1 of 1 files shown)


packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -8,14 +8,14 @@ import { Blockchain, UI_RPC_METHOD_SOLANA_SIMULATE } from "@coral-xyz/common";
 import {
   useBackgroundClient,
   useBlockchainNativeTokens,
-  useBlockchainTokensSorted,
```
hkirat, line 11: Lingering unused import from https://github.com/coral-xyz/backpack/pull/1028/files

PR #1032, opened 2022-10-07
Update web3.js, add versioned transaction

hkirat:
The PR 
1. Upgrades the `web3.js` to the latest version to include newly introduced `version` and `legacy` tranasactions
2. Adds logic to sign `versioned` transactions.

Just tried [this](https://github.com/solana-labs/wallet-adapter/blob/master/packages/starter/example/src/components/SendV0Transaction.tsx) demo locally. Was able to send a V0 transaction [here](https://explorer.solana.com/tx/531gwphCXnQp63jxu4z9JKeMv5ECxjVG1pBSAYgowgNMZL2LeZvbfkEveEyG1HGjNvWRjDtn8dLvSeqMEVqLCfn?cluster=devnet)

Testing checklist - 
- [x] Legacy transactions
- [x] V0 Transactions
- [x] Old transaction type
- [x] Simulate Transactions
- [x] Legacy transactions in xnfts
- [x] V0 transactions in xnfts
- [x] Gas estimates in old 
- [ ] Gas estimation in the new flow


packages/provider-core/src/common/solana.ts
```diff
@@ -107,11 +121,14 @@ export async function signAllTransactions(
...
   requestManager: RequestManager,
   connection: Connection,
-  txs: Array<Transaction>
-): Promise<Array<Transaction>> {
+  txs: Array<Transaction | VersionedTransaction>
+): Promise<Array<Transaction | VersionedTransaction>> {
   let _blockhash: string | undefined;
   for (let k = 0; k < txs.length; k += 1) {
     const tx = txs[k];
+    if ("version" in tx) {
```
hkirat, line 129: Have a bunch of these in the codebase instead of a function to keep `typescript` happy (TS knows the tx is a `VersionTransaction` if we do a check like this, but it won't if we create a function that returns true or false. 
Anyone have any suggestions around a cleaner way to do this?

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -192,11 +196,16 @@ export function useSolanaTxData(serializedTx: any): TransactionData {
...
   useEffect(() => {
     const estimateTxFee = async () => {
-      const transaction = Transaction.from(bs58.decode(serializedTx));
+      const transaction = TransactionV2.from(serializedTx);
       let fee;
       try {
-        fee = await transaction.getEstimatedFee(connection);
-      } catch {
+        if (TransactionV2.isVersioned(transaction)) {
+          // TODO: Add gas estimation
```
hkirat, line 203: Gas estimation was merged in recently for the new transaction format - https://github.com/solana-labs/solana/pull/28217/files
No client side API has been added yet so we might have to either write a fn similar to https://github.com/solana-labs/solana-web3.js/blob/master/src/transaction/legacy.ts#L552 (more specificallly the `compileMessage function` in our codebase or contribute this back to `web3.js` and use it directly from there.

hkirat:
We have the following fns ported to use the new transaction formats - 
`send`
`signTransaction`
`signAllTransactions`
`simluateTransaction`
 (all the fns used in https://github.com/solana-labs/wallet-adapter/blob/af3dfedf9713c52a8f17809e344ecc3ac37b085f/packages/wallets/backpack/src/adapter.ts)
We still need to port find a way to serialize the new `MessageV0` to send it over the `getFeeForMessage` call and get the fee for a given transaction. 
This can be a part of this PR or a follow up based on how we would wanna proceed on https://github.com/coral-xyz/backpack/pull/1032#discussion_r990458680

packages/background/src/backend/core.ts
```diff
@@ -129,13 +131,23 @@ export class Backend {
 
   async solanaSimulate(
     txStr: string,
-    _walletAddress: string,
+    walletAddress: string,
     includeAccounts?: boolean | Array<string>
   ): Promise<any> {
-    const tx = Transaction.from(bs58.decode(txStr));
+    const tx = TransactionV2.from(txStr);
+    const signersOrConf =
```
hkirat, line 138: Have taken a call of always sending the users account if this is a `VersionedTransaction`. 
I think there could be a cleanup here to change the API from the top so we allow users to pass in either `signers` or `transactionSimulationConfig` eventually.

packages/provider-core/src/common/solana.ts
```diff
@@ -48,27 +50,34 @@ export async function sendAndConfirm(
...
+    if (!tx.recentBlockhash) {
+      const { blockhash } = await connection!.getLatestBlockhash(
+        options?.preflightCommitment
+      );
+      tx.recentBlockhash = blockhash;
+    }
+  } else {
+    if (signers) {
+      tx.sign(signers);
+    }
```
armaniferrante, line 80: Do we need to do a blockhash and fee payer check here?

packages/common/src/solana/TransactionV2.ts
```diff
@@ -0,0 +1,29 @@
...
+  static getSerializedMessage(txStr: string) {
+    const tx = this.from(txStr);
+    const versioned = "version" in tx;
+    return !versioned
+      ? Transaction.from(bs58.decode(txStr)).serializeMessage()
+      : VersionedTransaction.deserialize(
+          bs58.decode(txStr)
+        ).message.serialize();
+  }
+}
```
jordansexton, line 29: Any reason to make this a static class instead of just exporting separate functions (e.g. `deserializeTransaction`, `isVersionedTransaction`, `getSerializedTransactionMessage`)?

It's a bit confusing to call `TransactionV2.from` and get a _not v2_ transaction, but also, VersionedTransaction doesn't mean it's a `0` versioned transaction, and these functions aren't logically dependent.

packages/common/src/solana/TransactionV2.ts
```diff
@@ -0,0 +1,29 @@
...
+    } catch (e) {
+      return VersionedTransaction.deserialize(bs58.decode(serializedTx));
+    }
+  }
+  static isVersioned(tx: Transaction | VersionedTransaction) {
+    if ("version" in tx) {
+      return true;
+    }
+    return false;
+  }
```
jordansexton, line 19:
```suggestion
  static isVersioned(tx: Transaction | VersionedTransaction): tx is VersionedTransaction {
    return "version" in tx;
  }
```

This way you get type narrowing when you call this.

packages/provider-core/src/common/solana.ts
```diff
@@ -48,27 +50,34 @@ export async function sendAndConfirm(
...
-  }
-  if (!tx.feePayer) {
-    tx.feePayer = publicKey;
-  }
-  if (!tx.recentBlockhash) {
-    const { blockhash } = await connection!.getLatestBlockhash(
-      options?.preflightCommitment
-    );
-    tx.recentBlockhash = blockhash;
+  const versioned = "version" in tx;
```
jordansexton, line 61: This should call `isVersionedTransaction` (with type narrowing).

packages/provider-core/src/common/solana.ts
jordansexton: VersionedTransaction must have these set.

packages/provider-core/src/common/solana.ts
```diff
@@ -80,18 +89,23 @@ export async function send(
...
-): Promise<Transaction> {
-  if (!tx.feePayer) {
-    tx.feePayer = publicKey;
-  }
-  if (!tx.recentBlockhash) {
-    const { blockhash } = await connection!.getLatestBlockhash();
-    tx.recentBlockhash = blockhash;
+  tx: T
+): Promise<T> {
+  const versioned = "version" in tx;
```
jordansexton, line 100: This should call `isVersionedTransaction` (with type narrowing).

packages/provider-core/src/common/solana.ts
```diff
@@ -107,11 +121,14 @@ export async function signAllTransactions(
   publicKey: PublicKey,
   requestManager: RequestManager,
   connection: Connection,
-  txs: Array<Transaction>
-): Promise<Array<Transaction>> {
+  txs: Array<Transaction | VersionedTransaction>
+): Promise<Array<Transaction | VersionedTransaction>> {
```
jordansexton, line 125: You may want the type of all these similar functions to be something like

`<T extends Transaction | VersionedTransaction>(tx: T): T {`

This doesn't _really_ matter to a dapp since it's going to lose typesafety, but this is consistent with WalletAdapter's types and will make sure your internal function logic checks out. Logically, you expect that if you call this with one type of transaction, you get the same back.

packages/provider-core/src/common/solana.ts
jordansexton, line 129: Probably just findAll to use the helper function.

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -192,11 +196,16 @@ export function useSolanaTxData(serializedTx: any): TransactionData {
...
+          fee = 5000;
+        } else {
+          fee = await (transaction as Transaction).getEstimatedFee(connection);
```
jordansexton, line 206: Narrowing the type with the helper means you can remove this kind of cast.

jordansexton (commented):
Made a quick pass at @armaniferrante's request!

packages/common/src/solana/TransactionV2.ts
hkirat: Agreed, changed it to a `transaction-helper` file with exported functions.

packages/recoil/src/hooks/useTransactionData.tsx
armaniferrante: We can do this separately. I think we want to just roll our own for now until web3.js has the functionality we need.

PR #1034, opened 2022-10-07
Build: Use default generated-config.ts, remove .sh prompt, fix xnft-cli build.

skynetcap:
The motivation for this PR is that when installing backpack fresh from source, it requires some additional steps that result in a mouth full of glass such as:

- Configuring `config.sh` / `generated-config.ts`
- Building `common` and/or `common-public`

This PR fixes the issues by:

1. Hard-coding a default generated.config.ts with default values.
2. Removing repetitive pop-up of config.sh files (at least on Windows/powershell) during build.
3. Builds `common` and `common-public` before `xnft-cli`, to prevent the `npx cp (iframe dist)` error that is common.

Using this branch, I am able to clone the repo (checkout the branch), and only `yarn install` and `yarn build` from the root directory is needed for a successful build.

Open to suggestions on how to further remove the glass eating from build.

### DIFF
--- a/packages/common-public/package.json
+++ b/packages/common-public/package.json
@@ -6,7 +6,7 @@
   "types": "dist/esm/index.d.ts",
   "scripts": {
     "start": "yarn build",
-    "build": "./scripts/config.sh && yarn lint && tsc && tsc-alias && tsc -p tsconfig.rn.json",
+    "build": "yarn lint && tsc && tsc-alias && tsc -p tsconfig.rn.json",
     "dev": "tsc --watch",
     "lint": "eslint ./src --ext .js,.jsx,.ts,.tsx"
   },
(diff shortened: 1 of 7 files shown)


packages/common-public/package.json
```diff
@@ -6,7 +6,7 @@
   "types": "dist/esm/index.d.ts",
   "scripts": {
     "start": "yarn build",
-    "build": "./scripts/config.sh && yarn lint && tsc && tsc-alias && tsc -p tsconfig.rn.json",
```
armaniferrante, line 9: As a general rule of thumb, we don't want to check in these generated files to the repo. Is there a problem with running `yarn build` first before getting started with the repo? Seems like we should fix that underlying problem if it exists rather than removing the pre-build step.

skynetcap: I'll see if there's a better approach that doesn't involve hard-coding the generated config.

armaniferrante:

Issue #1050, opened 2022-10-09
Legacy transactions can be deserialized with VersionedTransaction.deserialize


jordansexton:
https://github.com/coral-xyz/backpack/blob/f720ff8a566b34f6b4cedd21742a500396d442ed/packages/common/src/solana/transaction-helpers.ts#L5-L12

It should be possible to only call `VersionedTransaction.deserialize(bs58.decode(serializedTx))`.

For legacy transactions deserializable as `Transaction`, this will produce a `VersionedTransaction` with `version` set to `'legacy'`. Calling code paths can just use `VersionedTransaction` everywhere.

Issue #1048, opened 2022-10-08
Add gas estimation in VersionedTransactions


hkirat:
More context in https://github.com/coral-xyz/backpack/pull/1032#discussion_r990458680 . 
Right now we've hardcoded it to 5000.

PR #1052, opened 2022-10-09
Follow ups versioned transactions

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/1050
Closes https://github.com/coral-xyz/backpack/issues/1048

The PR makes all transactions `Versioned` in the codebase (except provider APIs since they need to conform the solana spec)

The PR also adds fee estimation for V0 transactions.

### DIFF
--- a/packages/background/src/backend/solana-connection.ts
+++ b/packages/background/src/backend/solana-connection.ts
@@ -4,6 +4,7 @@ import {
   GetAccountInfoConfig,
   PublicKey,
   SimulateTransactionConfig,
+  VersionedMessage,
 } from "@solana/web3.js";
 import type {
   Commitment,
@@ -69,7 +70,7 @@ import type {
   BlockheightBasedTransactionConfirmationStrategy,
 } from "@solana/web3.js";
 import type { Notification, EventEmitter } from "@coral-xyz/common";
-import { decode } from "bs58";
+import { encode } from "bs58";
 import {
   getLogger,
   customSplTokenAccounts,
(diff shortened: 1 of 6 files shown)


packages/background/src/backend/solana-connection.ts
```diff
@@ -459,10 +460,20 @@ export class SolanaConnectionBackend {
...
   async getFeeForMessage(
-    message: Message,
+    message: VersionedMessage,
     commitment?: Commitment
   ): Promise<RpcResponseAndContext<number>> {
-    return await this.connection!.getFeeForMessage(message, commitment);
+    const encodedMessage = Buffer.from(message.serialize()).toString("base64");
+    return await this.connection!.getFeeForMessage(
+      {
+        serialize: () => ({
```
hkirat, line 469: This is ugly but only way to make it work until this API is fixed in `solana/web3.js` - https://github.com/solana-labs/solana-web3.js/blob/master/src/connection.ts#L3727

Right now it expects a `Message` but it should be changed to accept `Message` and `MessageV0`

hkirat: More context - If we pass in the message as if as the first argument, `message.serialize().toString("base64")` fails because `MessageV0.serialize()` returns a `UIntArray` and not a `Buffer` (unlike LegacyMessage which returns a Buffer)

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -200,12 +199,11 @@ export function useSolanaTxData(serializedTx: any): TransactionData {
...
       let fee;
       try {
-        if (isVersionedTransaction(transaction)) {
-          // TODO: Add gas estimation
-          fee = 5000;
-        } else {
-          fee = await transaction.getEstimatedFee(connection);
-        }
+        // TODO: Remove type inference after the API for `getFeeForMessage` changes
+        const response = await connection.getFeeForMessage(
```
hkirat, line 203: This will only start working for `TransactionV0` when https://github.com/solana-labs/solana/pull/28217/files reaches the respective clusters (It isn't yet deployed on either testnet or devnet).
Will work as before for `LegacyTransaction`s

PR #1059, opened 2022-10-10
adds nvmrc for v18

peterpme:
.nvmrc works with fnm and nvm


wentokay:
it might be worth adding something like this in the root package.json too for people who aren't using a version manager

```
{
  "engines": {
    "node": "18.x"
  }
}
```

wentokay:
maybe a yarn major version in engines too, I'm using 1.22.19 atm. (upgrading the major version or moving to pnpm one day could be nice)

it might not be necessary for the engines stuff though, wdyt @peterpme?

peterpme:
Great idea!

.github/workflows/pull_request.yml
```diff
@@ -275,7 +275,7 @@ jobs:
       # NPM Publish.
       ##########################################################################
 
-      - name:  "update npm package version"
+      - name: "update npm package version"
```
peterpme, line 278: random space heh

PR #1063, opened 2022-10-11
encode pass for memoryrouter during onboarding

### LINKED ISSUE #1062: passwords in onboarding URLs not escaped
Configure a wallet with the password `Test#123` and it is passed as is in the URL, making the routing think there is a fragment which results in a blank page after the create password step in onboarding.

wentokay:
fixes #1062


web/pages/privacy.tsx
```diff
@@ -0,0 +1,18 @@
+export default function Terms() {
```
armaniferrante, line 1: is this and the privacy pdf supposed to be there?

Issue #1045, opened 2022-10-08
invalid DOM nesting warning in development


tomlinton:
<img width="899" alt="Screen Shot 2022-10-08 at 12 12 56 PM" src="https://user-images.githubusercontent.com/489202/194691656-f42974b1-5959-42fe-b384-bd06121bee0d.png">

gala25:
I'd like to have a stab at it 🙂

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1066

PR #1066, opened 2022-10-11
Gala/fix dom nesting

gala25:
This fix implements the addition of the props attribute to the MuiButton call, so we can pick a different component for the button, to avoid button nesting which caused the error. In addition the `onClick` handler has been moved to the top level element as per recommendations.


gala25:
Fixes #1045

Issue #1044, opened 2022-10-08
this.isConnected is not a function in ethereum provider


tomlinton:
Open example client, connect Ethereum wallet, refresh the page. Possibly due to object freezing/private members in https://github.com/coral-xyz/backpack/pull/925? Have not investigated.

<img width="885" alt="Screen Shot 2022-10-08 at 6 56 53 PM" src="https://user-images.githubusercontent.com/489202/194691625-bca84329-570f-4d6c-802e-c89c8acc4d89.png">

PR #1067, opened 2022-10-11
Make all fns in ethereum-provider arrow fns

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/1044
The PR converts `handleEthAccounts` to an arrow fn as it was being bound to this before https://github.com/coral-xyz/backpack/pull/925 . 

Wasn't able to reproduce the error locally but I think this should fix it. @tomlinton could you help me confirm the fix if you can still recreate this locally?

### DIFF
--- a/packages/provider-core/src/provider-ethereum.ts
+++ b/packages/provider-core/src/provider-ethereum.ts
@@ -425,12 +425,12 @@ export class ProviderEthereumInjection extends PrivateEventEmitter {
   /**
    * Handle eth_accounts requests
    */
-  async #handleEthAccounts() {
+  #handleEthAccounts = async () => {
     if (this.isConnected() && this.publicKey) {
       return [this.publicKey];
     }
     return [];
-  }
+  };
 
   /**
    * Handle eth_requestAccounts requests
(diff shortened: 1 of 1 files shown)


packages/provider-core/src/provider-ethereum.ts
```diff
@@ -425,12 +425,12 @@ export class ProviderEthereumInjection extends PrivateEventEmitter {
   /**
    * Handle eth_accounts requests
    */
-  async #handleEthAccounts() {
```
armaniferrante, line 428: If we need to do this do we need to make everything else arrow functions? E.g. `handleAccountsChanged`?

hkirat: No harm in doing that. Right now I kept it backward compatible with the list [here](https://github.com/coral-xyz/backpack/pull/925/files#diff-62fce178271dfd7972ea13095782602544cec13ad12014330d08245bf4d75bdaL146). 

I'll make all of em arrow.

hkirat:
Any objections to converting all of the fns in all providers to arrow fns?
No real harm in doing it, no immediate benefit either unless fns inside them are called out of context

tomlinton:
No objections, this seems to fix it from my local testing. I feel like prior to this we had autoreconnect working here. I might be wrong, can deal with it as a separate issue anyway.

Here is the repro if you want to try it:

https://user-images.githubusercontent.com/489202/195196421-fa3495ce-70b5-4107-80aa-c46434186115.mov

tomlinton (approved):

Issue #1073, opened 2022-10-11
make it possible to onboard with any blockchain


tomlinton:
Right now onboarding is limited to Solana only.

Issue #991, opened 2022-10-03
inject backpack version identifier into User-Agent request header


wentokay:
this could be done automatically by intercepting outgoing requests with a service worker

it would be useful for determining what response to serve from different backend services if we need to have versioning or quickly preparing staging environments

PR #1077, opened 2022-10-12
add blockchain selector to onboarding

tomlinton:
- Adds a blockchain selector to the onboarding flow to allow users to onboard with any supported blockchain.
- Updates the backend and worker code to allow the above

Screen Shot 2022-10-14 at 10 24 08 AM

Closes https://github.com/coral-xyz/backpack/issues/1073


packages/background/src/backend/core.ts
```diff
@@ -402,6 +404,7 @@ export class Backend {
           inviteCode,
           publicKey,
           waitlistId,
+          blockchain,
```
tomlinton, line 407: Handle this in the signature verification on the backend

packages/app-extension/src/components/Onboarding/pages/RecoverAccount.tsx
```diff
@@ -41,7 +41,10 @@ export const RecoverAccount = () => {
         if (!res.ok) throw new Error(json.message);
 
         navigate(
-          `${pathname}/${JSON.stringify({ username, pubkey: json.pubkey })}`
+          `${pathname}/${json.blockchain}/${JSON.stringify({
```
tomlinton, line 44: blockchain should be returned in the json response from the server for use in the recovery flow here

packages/app-extension/src/components/Onboarding/pages/ImportAccounts.tsx
```diff
@@ -7,12 +7,16 @@ import {
...
+    blockchain: Blockchain;
+    mnemonic: string;
+  }>();
+
   const navigate = useNavigate();
   return (
     <div>
       <ImportAccounts_
-        blockchain={Blockchain.SOLANA}
+        blockchain={blockchain!}
```
tomlinton, line 19: We should default to Blockchain.SOLANA if its not in the params as it might be an old build

tomlinton: Actually that shouldn't be possible

tomlinton:
@wentokay this is ready but would like to get your sign off on it. Also not clear on how we deploy and run the migration as I understand it isn't automated.

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
```diff
@@ -0,0 +1,394 @@
...
+    >
+      <Box>
+        <Box
+          sx={{
+            marginLeft: "24px",
+            marginRight: "24px",
+            marginTop: "24px",
+          }}
+        >
+          <Header text="Whch network would you like Backpack to use?" />
```
wentokay, line 28: *Which

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
```diff
@@ -0,0 +1,394 @@
...
+          </SubtextParagraph>
+        </Box>
+      </Box>
+      <Box style={{ padding: "0 16px 16px" }}>
+        <Grid container spacing={1.5}>
+          <Grid item xs={6}>
+            <ActionCard
+              icon={<EthereumIcon />}
+              text={"Ethereum"}
+              onClick={() => onSelect("ethereum" as Blockchain)}
```
wentokay, line 41: could this be Blockchain.ETHEREUM insead?

wentokay:
not sure if I've done something wrong yet but I don't think the import flow is working? It's generating a new mnemonic
https://user-images.githubusercontent.com/101902546/195736347-52fd5476-47bb-45d6-819c-137ae7b1eb5f.mov

tomlinton:
> not sure if I've done something wrong yet but I don't think the import flow is working? It's generating a new mnemonic
> 
>  import.mov

Looks like its hitting the wrong route

backend/workers/auth/src/index.ts
```diff
@@ -141,13 +152,22 @@ app.post("/users", async (c) => {
...
-  if (
-    !sign.detached.verify(
+  let isValidSignature = false;
+  if (variables.blockchain === "solana") {
+    isValidSignature = sign.detached.verify(
       Buffer.from(JSON.stringify(body), "utf8"),
       decode(c.req.header("x-backpack-signature")),
       decode(variables.publicKey)
-    )
-  ) {
```
wentokay, line 150: I think this might break backwards compatibility with the existing extension(s) deployed in the chrome store, v0.2.0 (and due to the way updates are pushed, there's a chance that a very small percentage users might still be running v0.1.0)

we'd probably need to set `variables.blockchain = "solana"` by default, so that when it's `undefined` the endpoint still works

it might be good to set up something [like this](https://github.com/coral-xyz/backpack/issues/991) soon, or have a dedicated endpoint for each release, i'm not sure what would be the best approach

wentokay: I added a default value [here](https://github.com/coral-xyz/backpack/blob/00a12760d9297e9158208675b7cbe64bb13c56fe/backend/workers/auth/src/index.ts#L165)

just running some tests now

PR #1078, opened 2022-10-12
Adds image proxying using imageproxy

hkirat:
The PR adds - 
1. Logic to proxy all image requests through our in-house imageproxy instance, we fallback to original if the request fails
2. Adds docker files and fly configuration file to the codebase that we should hook up to CI/CD (Right now I've manually deployed to fly)

TODO: Add a CDN in front of the proxy to further cache URLs


armaniferrante:
LGTM. Let's create an issue for the CDN work.

Issue #1081, opened 2022-10-12
window.xnft object unable to sign transactions after changing urls


armaniferrante:
This is the case for hash routers as well.

To summarize:

- use iframe
- window.xnft works
- change url (inside same iframe)
- window.xnft no longer works

Report: https://discord.com/channels/985994296337498182/1004775994193875064/1029756240437448774

dfl-zeke:
@armaniferrante
So after digging for a while, I've found where this problem occurs, it's on this line https://github.com/coral-xyz/backpack/blob/cf0535c11ea4341b295d34625a7023bc27234220/packages/common/src/channel/plugin.ts#L29
 
 I don't know logic behind this comparison, so I'll have to push more to detect why they mismatch. Easy fix is to write this instead of that comparison `!event.data.href.includes(url.href)` but I don't think it will fix for all cases.

dfl-zeke:
Close this issue, it's fixed.

PR #1087, opened 2022-10-12
window.xnft object unable to sign transactions after changing urls fix.

dfl-zeke:
This commit fixes issue (#1081 ). I tried to fix it by saving url, in case parent object is true. Maybe this is not the best solution for other cases.

### DIFF
--- a/packages/provider-core/src/request-manager.ts
+++ b/packages/provider-core/src/request-manager.ts
@@ -14,6 +14,7 @@ export class RequestManager {
   private _requestChannel: string;
   private _responseChannel: string;
   private _parent?: boolean;
+  private _url?: string;
 
   constructor(
     requestChannel: string,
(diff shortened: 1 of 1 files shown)


packages/provider-core/src/request-manager.ts
```diff
@@ -63,7 +67,7 @@ export class RequestManager {
       window.parent.postMessage(
         {
           type: this._requestChannel,
-          href: window.location.href,
+          href: this._url ?? window.location.href,
```
armaniferrante, line 70:
```suggestion
          // this._url will always be set here, because this._parent is true.
          href: this._url!,
```

armaniferrante: Haven't tested this, but I expect it to be correct.

dfl-zeke:
So far, it works perfectly. I'll watch if I encounter any issue.

armaniferrante:
> So far, it works perfectly. I'll watch if I encounter any issue.

To clarify, did you test my suggested change?

dfl-zeke:
No, I couldn't find your comment because I had adlocker on and could not see it... I will test it again with these changes you suggested and I guess it will work the same, but let's test first.

dfl-zeke:
It works as it should! I think it's ready to commit.

PR #1090, opened 2022-10-12
 New way of rendering

hkirat:
The PR adds a new way of securely rendering xnfts. 
The new model renders everything inside the iframe by injecting the renderer inside the xnft. 
Things to fix
 - [x] Small flash of white before dark theme pops up in iframe
 - [x] Figure out backward compatibility with currently deployed xnfts, letting the developers know they'll all have to update since this is a breaking change.
 - [x] Pass off iframe context if there is a child iframe
 - [x] Show warnings to existing xnfts

Testing checklist
 - [x] All examples
 - [x] All clients
 - [x] An existing deployed xnft - show warning when this happens
 - [x] App Extension
 - [x] New Prices xnft


hkirat:
This should be ready for review. 
I'm still to figure out the best way to show a warning for broken xnfts.
Right now it's just a white iframe.

Have tested all existing examples, clients.

Few things that have changed - 
1. All Iframes now send and receive RPC events and events like `mount`, `connect` from the `SolanaXnftInjector`.
2. Components are not shared b/w the `xnft renderer` and the `extension` anymore. This'll give us more freedom to chose different frameworks (tailwind, material ui) and different react versions.

hkirat:
Added another PR forked out of the base of this PR - https://github.com/hkirat/backpack/pull/1 . This PR adds logic to render a warning when an existing xnft is accessed in the new renderer.
Have kept it separate since this is already very big to review, will merge that in independently

packages/react-xnft/src/reconciler.ts
```diff
@@ -700,20 +596,13 @@ function createImageInstance(
   _o: OpaqueHandle
 ): ImageNodeSerialized {
   const id = h.nextId();
-  let onClick = false;
```
armaniferrante, line 703: Note to self: This is no longer needed because the click handlers are directly registered on the components as events are no longer being bridged to the host.

packages/react-xnft/src/ReactDom.ts
```diff
@@ -0,0 +1,281 @@
...
+} from "./index";
+
+import { getLogger } from "@coral-xyz/common-public";
+const logger = getLogger("react-xnft/reconciler");
+
+//
+// Note that we only handle methods in the "commit" phase of the react
+// reconciler API.
+//
+export class ReactDom {
```
armaniferrante, line 16: Note to self: This used to be on the other side of the bridge (in the host) but now lives inside the iframe. This keeps references to all components in the tree and rerenders them when needed.

packages/react-xnft/src/reconciler.ts
hkirat: We might need to add these back though for mobile since over there we'll still be talking over postmessage

packages/react-xnft/src/reconciler.ts
```diff
@@ -22,21 +10,7 @@ const events = new EventEmitter();
 export const ReactXnft = {
   events,
   render(reactNode: any) {
-    window.onload = () => {
-      window.xnft.on("click", (event: Event) => {
```
armaniferrante, line 26: Note to self: these event handlers are no longer needed because we no longer bridge events across to the iframe. Event handlers are directly registered on the components now.

packages/common-public/src/constants.ts
```diff
@@ -1,30 +1,5 @@
...
-export const RECONCILER_BRIDGE_METHOD_INSERT_BEFORE =
-  "reconciler-bridge-method-insert-before";
-export const RECONCILER_BRIDGE_METHOD_REMOVE_CHILD =
-  "reconciler-bridge-method-remove-child";
-export const RECONCILER_BRIDGE_METHOD_REMOVE_CHILD_FROM_CONTAINER =
-  "reconciler-bridge-method-remove-child-from-container";
-export const RECONCILER_BRIDGE_METHOD_RENDER_ROOT =
-  "reconciler-bridge-method-render-root";
-
 export const CHANNEL_PLUGIN_REACT_RECONCILER_BRIDGE =
```
armaniferrante, line 1: Is this channel used anywhere still?

packages/app-extension/src/plugin/Component.tsx
```diff
@@ -1,10 +1,5 @@
-import React, { useState } from "react";
```
armaniferrante, line 1: Note to self: The reason why this file is not shared with `react-xnft-dom-renderer` is so that we can decouple the react-xnft framework styling from the app-extension, e.g., so that we can use tailwind or something else.

packages/app-extension/src/plugin/Component.tsx
```diff
@@ -208,40 +207,9 @@ const useStyles = styles((theme) => ({
   },
 }));
 
-export function Component({ viewData }) {
+export function Component({ viewData }: any) {
```
armaniferrante, line 210: Is this ever used anymore inside the app-extension context? Probably not.

hkirat: It is used by the `ViewRenderer` which is in turn used by classes like `BalancesTableFooter` inside `plugin/Component.tsx`

armaniferrante: BalancesTableFooter can be removed. Let's see if we can remove `Component` from here.

armaniferrante: By extension we cal also remove ViewRenderer from app-extension.

hkirat: Yup that makes sense, removed `ViewRenderer` and `Component` from the `app-extension`

packages/app-extension/src/plugin/Component.tsx
```diff
@@ -0,0 +1,275 @@
+import { TextareaAutosize as MuiTextArea } from "@mui/base";
```
armaniferrante, line 1: There's probably a better path for this to live.

packages/common/src/plugin.ts
```diff
@@ -69,23 +52,21 @@ export class Plugin {
   private _activeWallets: { [blockchain: string]: string };
   private _connectionUrls: { [blockchain: string]: string | null };
   private _rpcServer: PluginServer;
-  private _bridgeServer: PluginServer;
-  private _iframeRoot?: HTMLIFrameElement;
+  public iframeRoot?: HTMLIFrameElement;
   private _iframeActive?: HTMLIFrameElement;
   private _nextRenderId?: number;
   private _pendingBridgeRequests?: Array<any>;
```
armaniferrante, line 58: We can get rid of this variable now?

packages/provider-core/src/provider-solana-xnft.ts
```diff
@@ -189,6 +193,30 @@ export class ProviderSolanaXnftInjection
     });
   }
 
+  public async addIframe(iframeEl) {
```
armaniferrante, line 196: What is this method used for?

packages/app-extension/src/plugin/Component.tsx
armaniferrante: `src/components/common` is probably better.

packages/provider-core/src/provider-solana-xnft.ts
hkirat: Whenever an iframe is put inside an xnft (or recursively an iframe inside an iframe inside an xnft), this method is called to ensure we pass all messages from the parent xnft down to all child iframes. 
This is what replaces this logic here - https://github.com/coral-xyz/backpack/blob/master/packages/react-xnft-renderer/src/plugin.ts#L203

Earlier we would make the deepest iframe active and disable all other iframes from receiving messsages. 
With this change all iframes are able to receive messages.

hkirat: It's called here - https://github.com/coral-xyz/backpack/pull/1090/files#diff-60ff5434f0c7e93bc972f52821f707c3e9e5995271cfe9a82c2c1b455e1dd39bR392 
Earlier we had a message here that we sent to `Plugin.ts` to make this iframe the active one.

armaniferrante: > I do think that we should rename this to `XnftProvider` and separate out the `solanaXnft` logic into another file.

Agreed.

hkirat: Filed https://github.com/coral-xyz/backpack/issues/1162

PR #1093, opened 2022-10-13
Fix first digit deletion on autolock input

### LINKED ISSUE #1074: First digit of Auto Lock Timer cannot be deleted with 'delete' key. must be highlighted then new digit typed
First digit of Auto Lock Timer cannot be deleted with 'delete' key. must be highlighted then new digit typed

Ex. 15 is default. Delete key will remove the 5 but get stuck in front of the 1. To change to '20' must highlight the numbers and type 20 on the keypad.

gala25:
This fix changes the TextField input type to text with a numeric input mask as per the recommendation on - [Text Field](https://mui.com/material-ui/react-text-field/)

Now the user can delete the first digit and the input will default to 0, while the greater than zero check will happen in the `onSet` function.

Fixes #1074


armaniferrante:
Thank you!

PR #1101, opened 2022-10-13
remove unnecessary asyncs in recoil selectors

tomlinton:


armaniferrante:
For my own understanding, is there a functional change here?

tomlinton:
No I don't believe so

PR #1102, opened 2022-10-13
allow navigator to set default tab

0xCactus:


packages/react-xnft/src/sdk/Tab.tsx
```diff
@@ -13,19 +13,22 @@ function Navigator({
   style,
   top,
   disableScroll,
-  options,
```
armaniferrante, line 16: This needs to be added back right?

packages/react-xnft/src/sdk/Tab.tsx
```diff
@@ -13,19 +13,22 @@ function Navigator({
...
+  initialTab,
```
armaniferrante, line 16: Please rename this to `initialRouteName` to be consistent with https://reactnavigation.org/docs/bottom-tab-navigator

packages/react-xnft/src/sdk/Tab.tsx
0xCactus: oops added it back

packages/react-xnft/src/sdk/Tab.tsx
0xCactus: done

PR #1147, opened 2022-10-17
Wallet Standard 1.0.0-rc.0

jordansexton:
This PR inlines the Wallet Standard implementation for Backpack into this repo so that Backpack owns the package for security. Care has been taken to add zero new runtime dependencies (only dev dependencies for types).

As before, `window.navigator.wallets` is added, referencing window.backpack but not altering it. There will be one more PR to update this package to 1.0.0 when the Wallet Standard packages exit RC.


packages/wallet-standard/src/endpoint.ts
```diff
@@ -0,0 +1,26 @@
+// This file is copied with modification from @solana/wallet-standard-util
+
+import type { SolanaChain } from './solana.js';
+import { SOLANA_DEVNET_CHAIN, SOLANA_LOCALNET_CHAIN, SOLANA_MAINNET_CHAIN, SOLANA_TESTNET_CHAIN } from './solana.js';
+
+export const MAINNET_ENDPOINT = 'https://solana-rpc-nodes.projectserum.com';
```
jordansexton, line 6: @armaniferrante should this be defined by an env var rather than hardcoded here?

jordansexton: Actually, that might be tricky. The build is only using tsc, so we can't interpolate env vars. If this is needed, I'll PR separately.

Issue #1111, opened 2022-10-14
allow .sol and .eth domains in send to fields


tomlinton:
Needs it for both token and NFT sends

Gajesh2007:
like people can send their USDC/SOL/ETH to a SNS or ENS address?

tomlinton:
Yea, the field validation prevents it right now. Should work without any further changes for Ethereum.  Not sure about Solana.

tomlinton:
Closed by #1156

PR #1156, opened 2022-10-17
implemented .eth & .sol transfers

Gajesh2007:
#1111 
Screenshot 2022-10-17 at 10 00 26 PM
Screenshot 2022-10-17 at 10 01 03 PM


packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx
```diff
@@ -588,14 +593,37 @@ export function useIsValidAddress(
...
         let pubkey;
-        try {
-          pubkey = new PublicKey(address);
-        } catch (err) {
-          setAddressError(true);
-          // Not valid address so don't bother validating it.
-          return;
+
+        // SNS Domain
+        if (address.includes(".sol")) {
```
armaniferrante, line 599: @johnrees do we have username validation for Backpack usernames to not include .sol or .eth?

wentokay: yep, periods aren't allowed

johnrees: yes, only a-z0-9_ is allowed

wentokay:
checking the yarn.lock PR action now, it's not the fault of this PR 🔬

wentokay:
added a related issue https://github.com/coral-xyz/backpack/issues/1157

packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx
Gajesh2007: there are also emojis right?

johnrees: https://github.com/coral-xyz/backpack/blob/87cf2fd79646f05546587ed1758336f2368cf1ad/backend/workers/auth/src/index.ts#L21-L22

packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx
```diff
@@ -582,14 +593,37 @@ export function useIsValidAddress(
...
+          try {
+            const hashedName = await getHashedName(address.replace(".sol", ""));
+            const nameAccountKey = await getNameAccountKey(
+              hashedName,
+              undefined,
+              new PublicKey("58PwtjSDuFHuUkYjH9BYnnQKHfwo9reZhC2zMJv9JPkx") // SOL TLD Authority
+            );
+            const owner = await NameRegistryState.retrieve(
+              new Connection(solanaConnection?.rpcEndpoint?.toString()),
```
tomlinton, line 608: just `solanaConnection` instead of `new Connection(solanaConnection?.rpcEndpoint?.toString())` here I think? If `solanaConnection` is undefined we should throw

tomlinton: By that a mean throw earlier than here because if we forget to pass the connection it'll look like a validation error 😬

packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx
```diff
@@ -615,16 +649,29 @@ export function useIsValidAddress(
...
         let checksumAddress;
-        try {
-          checksumAddress = ethers.utils.getAddress(address);
-        } catch (e) {
-          setAddressError(true);
-          return;
+
+        if (address.includes(".eth")) {
+          try {
+            checksumAddress = await ethereumProvider?.resolveName(address);
```
tomlinton, line 659: Similar to above, we should throw if ethereumProvider is undefined outside of this try/catch so it doesn't manifest as a validation error if we forget to pass it

tomlinton (commented):
https://github.com/coral-xyz/backpack/blob/ff7e4c6cfa9affb20dabd43a58cd1bbd59c3121a/packages/app-extension/src/components/Unlocked/Nfts/Detail.tsx#L196-L204 will need updating too

Gajesh2007:
on it

Gajesh2007:
done.

tomlinton:
Looks good! Just the merge conflict.

Gajesh2007:
✅

PR #1159, opened 2022-10-17
only allow 0x eth addresses in auth worker

### LINKED ISSUE #1158: ethers.utils.getAddress doesn't guarantee provided eth address is in 0x format
`ethers.utils.getAddress` won't throw an error in certain cases when the address doesn't start with `0x`

`getAddress("XE65GB6LDNXYOFTX0NSV3FUWKOWIXAMJK36");` is valid for example

if `XE65GB6LDNXYOFTX0NSV3FUWKOWIXAMJK36` is sent to the worker API we either need to reject the request as an invalid payload (preferred) or modify the payload and store the 0x format

https://docs.ethers.io/v5/api/utils/address/#utils-getAddress

https://github.com/coral-xyz/backpack/blob/87cf2fd79646f05546587ed1758336f2368cf1ad/backend/workers/auth/src/index.ts#L36

wentokay:
fixes #1158

### DIFF
--- a/backend/workers/auth/src/index.ts
+++ b/backend/workers/auth/src/index.ts
@@ -33,8 +33,12 @@ const BaseCreateUser = z.object({
 const CreateEthereumUser = BaseCreateUser.extend({
   publicKey: z.string().refine((str) => {
     try {
-      ethers.utils.getAddress(str);
-      return true;
+      if (
+        str === ethers.utils.getAddress(str) &&
+        str.startsWith("0x") // one extra check in case getAddress API changes
+      ) {
+        return true;
+      }
     } catch (err) {}
     return false;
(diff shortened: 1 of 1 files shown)


tomlinton:
Can we just use the return of getAddress? That will be the he checksummed version in a consistent format

wentokay:
> Can we just use the return of getAddress? That will be the he checksummed version in a consistent format

yeah sure I'm afk so can you make a commit in this branch if you think that makes more sense?

My reasoning for this way was it might be easier for us if we can make assumption that we're only ever handling 0x strings in the frontend & backend, the second conditional check was in case they changed the way the function worked in the future

armaniferrante:

Issue #1165, opened 2022-10-17
react-xnft: text field doesn't update


armaniferrante:
see the prices xnft

PR #1166, opened 2022-10-17
Fixes onChange event handlers on Input boxes

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/1165

Change API to `e.target.value` vs `e.data.value`. Added this change to the migration doc that'll be shared with devs


hkirat:
Prices example - 
Screenshot 2022-10-18 at 2 08 04 AM

PR #1023, opened 2022-10-05
Custom inputs for updating RPC URLs

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/941 

Awaiting review from design. 

### Solana
Screenshot 2022-10-06 at 2 18 12 AM
Screenshot 2022-10-06 at 2 37 08 AM
Screenshot 2022-10-06 at 2 37 15 AM
Screenshot 2022-10-06 at 2 37 25 AM

### Eth
Screenshot 2022-10-06 at 2 37 37 AM

Screenshot 2022-10-06 at 2 37 54 AM
Screenshot 2022-10-06 at 2 37 43 AM

PR #1170, opened 2022-10-17
eslint sort import configuration

tomlinton:
Can run this and merge the results whenever. Will wait for #1023 as it is large.

### DIFF
--- a/packages/app-extension/.eslintrc.js
+++ b/packages/app-extension/.eslintrc.js
@@ -23,7 +23,25 @@ module.exports = {
     "no-undef": IGNORE,
     "no-unused-vars": IGNORE,
     "prefer-const": IGNORE,
-    "simple-import-sort/imports": ERROR,
+    "simple-import-sort/imports": [
+      "error",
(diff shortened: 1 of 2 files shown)


wentokay:
oh nice, i _started_ doing some work locally on this earlier with a global eslint config package that'd be shared across the monorepo https://github.com/coral-xyz/backpack/commit/bae4c39bf8645b70d9d17c06f9cf2c039bfac704 as per https://turborepo.org/docs/handbook/linting/eslint

are you happy with these rules now? I could include them in the global config and we could hit everything in one go.. after any big PR merges like you thoughtfully mention!

tomlinton:
Global is even better. Those rules seem to work pretty well, as it groups imports in a sensible kind of way. We might need to exclude the `wallet-standard` package since it is inlined here and we don't want to make it hard to merge future changes.

tomlinton:
Closing this in favour of the global option!

wentokay:
superseded by #1180

Issue #1167, opened 2022-10-17
Fix fonts, no scrolling in xnft viewer


hkirat:
Broke recently in https://github.com/coral-xyz/backpack/pull/1090

PR #1172, opened 2022-10-17
Fixes fonts and scroll in xnfts

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/1167


hkirat:
Have taken a call to upload the fonts to cloudfront and use them here. 
Reason - If we import the fonts (like we do in app-extension), spit out a bunch of files (along with index.js) that we'll need to copy over to `xnft-cli`.

PR #1210, opened 2022-10-19
Update wallet-standard package

jordansexton:
This PR

- Fixes some lint issues
- Fixes the type of `sendAndConfirmTransaction` to support passing `VersionedTransaction` objects
- Makes the injection of all window objects robust against failures

The last point means that even if another extension overwrites window.backpack accidentally or maliciously, the Wallet Standard wrapper will still work.


packages/provider-injection/src/index.ts
```diff
@@ -22,21 +22,27 @@ function main() {
...
+    );
+  }
+
+  try {
+    Object.defineProperty(window, "ethereum", { value: ethereum });
+  } catch (e) {
+    console.warn(
+      "Backpack couldn't override window.ethereum, disable other Ethereum wallets to use Backpack"
+    );
+  }
```
jordansexton, line 42: We wrap each attempt in a separate guard so individual failures are tolerated better.

packages/provider-injection/src/index.ts
```diff
@@ -52,10 +58,14 @@ function initProvider() {
...
-  });
+    });
+  } catch (e) {
+    console.warn(
+      "Backpack couldn't override window.xnft, disable other xNFT wallets to use Backpack"
+    );
+  }
 
-  register();
+  register(solana);
```
jordansexton, line 68: Instead of relying on window.backpack inside of the Wallet Standard wrapper, we pass a reference here.

packages/wallet-standard/package.json
```diff
@@ -21,6 +21,7 @@
         "lint:fix": "yarn run fmt && eslint --fix ."
     },
     "dependencies": {
+        "@coral-xyz/provider-core": "*",
```
jordansexton, line 24: Using the `ProviderSolanaInjection` directly lets us remove some code and prevents issues where the expected types could diverge.

packages/wallet-standard/src/wallet.ts
```diff
@@ -106,23 +103,25 @@ export class BackpackWallet implements Wallet {
...
         }
 
-        window.backpack.on('connect', this.#connected);
-        window.backpack.on('disconnect', this.#disconnected);
-        window.backpack.on('connectionDidChange', this.#reconnected);
+        this.#backpack = backpack;
+
+        backpack.on('connect', this.#connected);
+        backpack.on('disconnect', this.#disconnected);
+        backpack.on('connectionDidChange', this.#reconnected);
```
jordansexton, line 115: Instead of accessing window.backpack, which could fail to have been written to the page, we use our private reference.

jordansexton:
> ## `yarn.lock` changes
> Click to toggle table visibility
> 
> Name 	Status 	Previous 	Current
> `react-xnft` 	[REMOVED](#) 	0.1.0-latest.2443 	-

I just ran `yarn install`, I assume this is okay?

wentokay:
> > ## `yarn.lock` changes
> > Click to toggle table visibility
> > Name 	Status 	Previous 	Current
> > `react-xnft` 	[REMOVED](#) 	0.1.0-latest.2443 	-
> 
> I just ran `yarn install`, I assume this is okay?

taking a quick look at what's happening now

there are some other strange things going on like "boring-avatars" being removed from the lockfile in this PR and the action not picking that up, our `yarn` setup needs some love atm

packages/provider-injection/src/index.ts
```diff
@@ -22,21 +22,27 @@ function main() {
...
+  try {
+    Object.defineProperty(window, "ethereum", { value: ethereum });
+  } catch (e) {
+    console.warn(
+      "Backpack couldn't override window.ethereum, disable other Ethereum wallets to use Backpack"
+    );
+  }
+
+  try {
+    Object.defineProperty(window, "backpack", {
```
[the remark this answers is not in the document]
jordansexton: Oops, thanks. Will fix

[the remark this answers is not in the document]
jordansexton: Fixed

PR #1216, opened 2022-10-19
general monorepo stability & maintainability updates

wentokay:
I've been trying to make yarn a bit more stable and reliable locally as issues with it, especially changes to the lockfile seem to be coming up quite frequently

In the end I started converting the monorepo to [pnpm](https://pnpm.io/) and during the process it revealed some issues which might affect us in future

instead of proposing to change the repo to pnpm immediately I have added some of the fixes that were undertaken in this PR

notably

- added missing dependencies in projects
- pinned the ledger package versions to the same versions that wallet-adapter uses, to avoid https://github.com/solana-labs/wallet-adapter/issues/499 which comes up if we did decide to switch package managers
- added a couple of ts-ignore and :any types to make the linter happy, will return to them just didn't want to get too off course rn


armaniferrante:

PR #1229, opened 2022-10-22
XnftMetadata to common-public

hkirat:
The PR moves `XnftMetadata` to common public to remove `@coral-xyz/common` dependency from `react-xnft`

### DIFF
--- a/packages/common-public/src/types.ts
+++ b/packages/common-public/src/types.ts
@@ -5,3 +5,9 @@ export type RpcRequest = {
 };
 
 export type Event = any;
+
+export interface XnftMetadata {
+  isDarkMode: boolean;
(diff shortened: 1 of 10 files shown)


armaniferrante:
I was thinking we might want to comine common + common-public? What's the advantage of getting rid of the common/ dependency?

hkirat:
@armaniferrante https://npmjs.com/package/@coral-xyz/common isn't public, and hence it began breaking for a few users using `ReactXnft` (not from the package, but from react-xnft-cli-quickstart). So I moved it out to unblock those users.

In the end yeah it makes sense to combine these since everything's public now!

armaniferrante:
Will make it public now.

PR #1238, opened 2022-10-23
Fixes iframe rpc messages

hkirat:
`CHANNEL_PLUGIN_RPC_REQUEST` request messages from child iframes inside an xnft are breaking.
This PR aims to fix that by introducing a `chain-request-manager` that chains requests and responses from a child iframe up to the `app-extension` and back to the child iframe

### DIFF
--- a/packages/provider-core/src/chained-request-manager.ts
+++ b/packages/provider-core/src/chained-request-manager.ts
@@ -1,0 +1,160 @@
+// A request manager that is used for chaining requests form
+// child IFRAMES to parent IFRAMES.
+
+import { getLogger, PluginServer } from "@coral-xyz/common";
+import type {
+  RpcRequest,
+  RpcResponse,
+  Event,
+  ResponseHandler,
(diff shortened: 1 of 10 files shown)


packages/provider-core/src/chained-request-manager.ts
```diff
@@ -0,0 +1,159 @@
...
+  }: RpcRequest): Promise<RpcResponse<T>> {
+    const id = this._requestId;
+    this._requestId += 1;
+
+    const [prom, resolve, reject] = this._addResponseResolver(id);
+    window.parent.postMessage(
+      {
+        type: this._requestChannel,
+        // this._url will always be set here, because this._parent is true.
+        href: this._url!,
```
armaniferrante, line 128: We will need to eventually assert this can't be spoofed in the future.

armaniferrante: Note this requirement did not originate from these changes.

PR #1242, opened 2022-10-24
Update Wallet Standard to latest Release Candidate

jordansexton:
This PR updates the Wallet Standard to the latest RC, which has some breaking API changes (mostly renaming interfaces).

It also sets up how the wallets attach to the window slightly differently. I'm marking the PR as a draft for now to test this with some apps first.


packages/wallet-standard/src/register.ts
```diff
@@ -1,13 +1,71 @@
...
+        console.error('wallet-standard:register-wallet event could not be dispatched\n', error);
+    }
+    try {
+        (window as WalletEventsWindow).addEventListener('wallet-standard:app-ready', ({ detail: api }) =>
+            callback(api)
+        );
+    } catch (error) {
+        console.error('wallet-standard:app-ready event listener could not be added\n', error);
+    }
+}
```
jordansexton, line 25: This is the focus of this update. We are no longer using the global `window.navigator.wallets` interface, but instead using the window's built-in event listener queue. This prevents wallets from directly manipulating the array before the app loads, and avoids conflicts with future window globals.

packages/wallet-standard/src/register.ts
```diff
@@ -1,13 +1,71 @@
...
+
+/** @deprecated */
+export function DEPRECATED_registerWallet(wallet: Wallet): void {
+    registerWallet(wallet);
+    try {
+        ((window as DEPRECATED_WalletsWindow).navigator.wallets ||= []).push(({ register }) => register(wallet));
+    } catch (error) {
+        console.error('window.navigator.wallets could not be pushed\n', error);
```
jordansexton, line 69: We preserve the existing behavior as well as the new behavior to avoid breaking any apps that have already updated to the latest RC of Wallet Adapter.

yarn.lock
```diff
@@ -1992,6 +1992,32 @@
...
+    "@ethersproject/properties" "^5.7.0"
+    "@ethersproject/random" "^5.7.0"
+    "@ethersproject/rlp" "^5.7.0"
+    "@ethersproject/sha2" "^5.7.0"
+    "@ethersproject/strings" "^5.7.0"
+    "@ethersproject/transactions" "^5.7.0"
+    "@ethersproject/web" "^5.7.0"
+    bech32 "1.1.4"
+    ws "7.4.6"
+
```
jordansexton, line 2020: Not sure why this is included here -- it's not a dependency of the changes in this PR. I'm guessing it's okay though, it's just Ethers.

jordansexton:
@armaniferrante this is good to go!

PR #1254, opened 2022-10-24
Allows whitelisted origins to make RPC calls to backpack

greenish:
Used to enable twitter authentication flow to reopen backpack when done.

### DIFF
--- a/packages/background/src/frontend/server-injected.ts
+++ b/packages/background/src/frontend/server-injected.ts
@@ -58,7 +58,10 @@ import { handlePopupUiResponse, RequestManager } from "./common";
 
 const logger = getLogger("server-injected");
 
-const whitelistedOrigins = ["http://localhost", "https://one-nft.vercel.app"];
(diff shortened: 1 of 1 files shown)


packages/background/src/frontend/server-injected.ts
```diff
@@ -145,7 +147,10 @@ async function handle<T = any>(
   ) {
     const origin = ctx.sender.origin;
     const isApproved = await ctx.backend.isApprovedOrigin(origin);
-    if (!isApproved) {
+    if (
+      !isApproved &&
+      !whitelistedOrigins.find((wlOrigin) => origin.startsWith(wlOrigin))
```
wentokay, line 152: you might want to get use `new URL(wlOrigin).hostname` or something like that

otherwise attacker could use https://one-nft.vercel.app.evilsite.com

greenish: Good point, updated to include `/` in the end so this isn't possible 👍

PR #1270, opened 2022-10-25
one to many pubkeys suggestions

wentokay:
additional suggestions for #1223

### DIFF
--- a/backend/reef/hasura/metadata/databases/default/tables/auth_publickeys_history.yaml
+++ b/backend/reef/hasura/metadata/databases/default/tables/auth_publickeys_history.yaml
@@ -8,4 +8,11 @@ insert_permissions:
       columns:
         - blockchain
         - publickey
+select_permissions:
+  - role: auth_worker
+    permission:
+      columns:
+        - blockchain
+        - publickey
         - user_id
+      filter: {}
--- a/backend/reef/hasura/metadata/databases/default/tables/auth_publickeys.yaml
+++ b/backend/reef/hasura/metadata/databases/default/tables/auth_publickeys.yaml
@@ -1,10 +1,3 @@
 table:
   name: publickeys
   schema: auth
-select_permissions:
-  - role: auth_worker
-    permission:
-      columns:
-        - blockchain
-        - publickey
-      filter: {}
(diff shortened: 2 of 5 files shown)


backend/reef/hasura/metadata/databases/default/tables/auth_publickeys_history.yaml
```diff
@@ -8,4 +8,11 @@ insert_permissions:
       columns:
         - blockchain
         - publickey
+select_permissions:
```
tomlinton, line 11: We shouldn't need this because we are querying the view `publickeys` not `publickeys_history` I think? Or does it need it for the view to work.

wentokay: needs it for the view to work, will explain in call

tomlinton:
Looks good, but how come the zeus files are removed?

wentokay:
I tweaked the zeus command so it has proper permissions for the auth_worker role, which meant a lot of the admin-only code and types got removed

added `--header=x-hasura-role:auth_worker`

backend/reef/hasura/metadata/databases/default/tables/auth_publickeys_history.yaml
tomlinton: Ahh and it works in local because I'm using some super admin permissions and not using the worker role?

tomlinton:
Oh awesome 👍

wentokay:
will leave it up to you to merge into your branch if/when you're ready :)

backend/reef/hasura/metadata/databases/default/tables/auth_publickeys_history.yaml
wentokay: exactly! now zeus is more strict it complained about accessing `publickeys` in the worker

you can quickly check what's allowed in the local hasura console (running by default on http://localhost:8112 although changes on that port won't get saved to files) by adding an extra `x-hasura-role` header and toggling it on and off

admin role | auth_worker role
------|------
Screen Shot 2022-10-25 at 20 57 12 |  Screen Shot 2022-10-25 at 20 57 18

ignore the content here, i'm on a different branch, just showing the difference in the amount of stuff in the left column

Issue #1244, opened 2022-10-24
Fix unstyled consistent scrollbars 


hkirat:
<img width="472" alt="Screenshot 2022-10-24 at 9 37 27 PM" src="https://user-images.githubusercontent.com/8079861/197573322-edba045a-160f-4100-bf9d-0b779b3c6828.png">

hkirat:
@armaniferrante could you confirm if this scrollbar is gone now from this `degods` page

PR #1276, opened 2022-10-26
Fixes scrollbars

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/1244
After trying out a bunch of things found https://github.com/xobotyi/react-scrollbars-custom that did the trick
We were originally using `react-custom-scrollbars` (very similar names amongst the two libraries). The old one has an issue on hover/if you have a connected mouse device where a white background shows up similar to https://github.com/coral-xyz/backpack/issues/1244 . 
This PR gets rid of those scrollbars.
Screenshot 2022-10-26 at 5 41 58 AM

Still need to decide how we change these scrollbars when the host theme changes. Right now it's an opaque black that works well on both light and dark but would be good to get design to help us figure out what the right thing to do here is.

### DIFF
--- a/packages/react-xnft-dom-renderer/package.json
+++ b/packages/react-xnft-dom-renderer/package.json
@@ -20,14 +20,15 @@
     "@emotion/styled": "^11.8.1",
     "@mui/base": "^5.0.0-alpha.101",
     "@mui/icons-material": "^5.8.4",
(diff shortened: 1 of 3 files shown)


hkirat:
Since this is a change in the renderer it can be deployed immediately (doesn't need to wait on app store)

Issue #1261, opened 2022-10-25
trim balance lengths in token table


tomlinton:
Although this is correct, it looks ugly.

<img width="373" alt="Screen Shot 2022-10-25 at 4 59 56 PM" src="https://user-images.githubusercontent.com/489202/197679256-8199a784-2444-4a03-af3c-dae81b9ab403.png">

Gajesh2007:
how short?

PR #1280, opened 2022-10-26
trimmed balance lengths in token table

Gajesh2007:
#1261 @tomlinton


yarn.lock
```diff
@@ -9560,6 +9560,11 @@ clsx@^1.1.1:
   resolved "https://registry.yarnpkg.com/clsx/-/clsx-1.1.1.tgz#98b3134f9abbdf23b2663491ace13c5c03a73188"
   integrity sha512-6/bPho624p3S2pMyvP5kKBPXnI3ufHLObBFCfgx+LkeR5lg2XYy2hqZqUf45ypD8COn2bhgGJSUE+l5dhNBieA==
 
+cnbuilder@^3.1.0:
+  version "3.1.0"
```
hkirat, line 9564: Can you revert the lock changes?

packages/app-extension/src/components/Unlocked/Balances/Balances.tsx
```diff
@@ -178,7 +178,7 @@ export function BalancesTableCell({ props, style }: any) {
         </div>
         <div className={classes.tokenListItemRow}>
           {subtitle && (
-            <Typography className={classes.tokenAmount}>{subtitle}</Typography>
+            <Typography className={classes.tokenAmount}>{subtitle.split(".")[0]}.{subtitle.split(".")[1].slice(0,5)}</Typography>
```
tomlinton, line 181: It might be worth handling the undefined case when the balance doesn't have decimals, e.g. `"12312312".split(".")[1]` might be undefined. Not sure if this is ever possible with the current blockchains but it might be with future blockchains.

Gajesh2007:
✅

Issue #1260, opened 2022-10-25
Apply useTheme() to components automatically


armaniferrante:
E.g., to text components

Benhawkins18:
Following

PR #1282, opened 2022-10-26
Added theming to Text, fixed colors of loader

hkirat:
In ref to https://github.com/coral-xyz/backpack/issues/1260

### DIFF
--- a/packages/react-xnft-dom-renderer/src/Component.tsx
+++ b/packages/react-xnft-dom-renderer/src/Component.tsx
@@ -207,6 +207,13 @@ const useStyles = styles((theme) => ({
     width: "18px",
     color: theme.custom.colors.secondary,
   },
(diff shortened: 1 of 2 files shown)


packages/react-xnft-dom-renderer/src/Component.tsx
```diff
@@ -946,12 +954,28 @@ export function __Button({
...
 
 function Loading({ id, props, style }: any) {
-  const theme = useCustomTheme();
+  const classes = useStyles();
   style = {
-    color: theme.custom.colors.activeNavButton,
     ...style,
   };
-  return <CircularProgress style={style} thickness={6} />;
+  return (
```
armaniferrante, line 961: Note that we probably want to be able to override the styles in the future.

PR #1288, opened 2022-10-26
Added tailwind support

hkirat:
Not every current component has been moved to TW (things like `BalancesTable`). 
Core components have been ported (Image, Svg, View, Text, Button)

Screenshot 2022-10-27 at 4 57 44 AM
Screenshot 2022-10-27 at 4 57 32 AM

We currently import tailwind from cdn (~300kb). There is scope for optimisations here where we can pick out classes present in bundles and create smaller css files per xnft.


examples/xnft/tailwind/src/app.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+function Metadata() {
+  const [inputText, setInputText] = useState("");
+  const metadata = useMetadata();
+  const connection = useConnection();
+  console.log("conn is " + connection);
+  return (
+    <View>
+      <View twClassName="p-2">
```
armaniferrante, line 26: Instead (or maybe in addition) of `twClassName` it would be nice to just say `tw` for short.

PR #1296, opened 2022-10-27
fix casing on publicKey in auth worker

tomlinton:


backend/workers/auth/src/index.ts
```diff
@@ -311,7 +311,7 @@ app.post("/users", async (c) => {
   if (c.env.SLACK_WEBHOOK_URL) {
     try {
       const publicKeyStr = publicKeys!
-        .map((b) => `${b.blockchain.substring(0, 3)}: ${b.publicKey}`)
+        .map((b) => `${b.blockchain.substring(0, 3)}: ${b.publickey}`)
```
armaniferrante, line 314: 🤔

wentokay: it's because `b.publicKey` is currently `undefined`, will come back to the whole [consistent casing stuff](https://github.com/coral-xyz/backpack/pull/1298#pullrequestreview-1158154763) another time

PR #1311, opened 2022-10-28
Support for Compression for NFTs

jnwng:
this is a preliminary take on integrating the Metaplex Read API to power the "collectibles" tab in Backpack. here's [more information](https://github.com/metaplex-foundation/api-specifications/tree/b9f0528d2f5ab0c7176699824e9e05db362324a4/specifications/read_api) about the Read API along with the specifications. 

critically, the Read API serves both [compressed](https://github.com/metaplex-foundation/metaplex-program-library/discussions/648) and uncompressed NFTs all in the same interface, which we're using here when a custom RPC supports the Read API.

transfers also work too!

one can try this out by turning on the "Compressed NFT" feature in the Solana settings
image

**review notes**
there's a lot of TODOs here because i have some questions for the Backpack team.
~~1. the Read API provides "pagination", but i don't think the Backpack interface does. should i go ahead and retrieve as much as possible + automatically paginate, or are there plans to provide pagination / filter controls?~~
~~2. some assets will be "compressed", and while the "read" flow is identical to uncompressed NFTs, the transfer is not. is it critical to surface that these assets are compressed in the UI, or just in the transfer process?~~

## remaining TODOs
- [ ] Currently only 5 assets are being retrieved. Plan on paginating through everything possible since this UI is grouped by collections... probably not optimal
- [ ] Need more error handling when proofs are out of date or stale data is in the interface


packages/recoil/src/atoms/solana/metaplex-read-api.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    let backpackNft: SplNftMetadataString & {
+      external_url: string;
+      image: string;
+      metadata: Omit<
+        SplNftMetadataString["metadata"],
+        // This is currently being omitted because they're not displayed in a meaningful way
+        "key" | "primarySaleHappened" | "isMutable" | "editionNonce"
+      >;
+    } = {
```
jnwng, line 91: a `backpackNft` declares a whole bunch of fields, but only a few are displayed. i've optimized for what is displayed and working back into all the rest of the types, but some of them are not useful because the Read API makes things easier (i.e., `uri` is only useful for retrieving an off-chain JSON manifest that the Read API is already surfacing)

jnwng:
this code is super rough, but i decided to send the PR to get early feedback rather than over-optimize. the Metaplex Read API in devnet is _also_ not 100% ready so this won't work (yet) out of the box, but the code can be evaluated beforehand while the final touches get sorted out on the API side

packages/recoil/src/atoms/solana/nft.tsx
```diff
@@ -21,15 +24,25 @@ export const solanaNftCollections = selector<NftCollection[]>({
     //
     const connectionUrl = get(solanaConnectionUrl)!;
     const publicKey = get(solanaPublicKey)!;
-    const { splNftMetadata: metadata } = get(
-      customSplTokenAccounts({ connectionUrl, publicKey })
-    );
+
+    let metadata: Map<string, SplNftMetadataString>;
+    if (BACKPACK_FEATURE_READ_API) {
```
armaniferrante, line 29: These two branches are the same but I suspect that's WIP

armaniferrante:
@jnwng on your two questions. I think the immediate solution (though something we should come back to) is to

1. Fetch everything you can and automatically paginate
2. Treat compressed and uncompressed as the same in the UI. I'm not married to this and would be interested in arguments against this.

packages/common/scripts/config.sh
```diff
@@ -41,6 +41,7 @@ export const BACKPACK_FEATURE_LIGHT_MODE = ${BACKPACK_FEATURE_LIGHT_MODE:=true};
 export const BACKPACK_FEATURE_POP_MODE = ${BACKPACK_FEATURE_POP_MODE:=true};
 export const BACKPACK_FEATURE_USERNAMES = ${BACKPACK_FEATURE_USERNAMES:=false};
 export const BACKPACK_FEATURE_XNFT = ${BACKPACK_FEATURE_XNFT:=true};
+export const BACKPACK_FEATURE_READ_API = "${BACKPACK_FEATURE_READ_API:=}";
```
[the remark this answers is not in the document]
jnwng: anything i need to take care of here? happy to update this if needed

packages/recoil/src/atoms/solana/nft.tsx
jnwng: this was intentional—wanted the code flow to be identical whether it was going through the Read API or the built-in Backpack RPC. that way we wouldn't need to much with the UI parts, just the data retrieval

jnwng:
> 1. Fetch everything you can and automatically paginate

added in automatic pagination. surfaced a couple of issues in the Read API from this front, but basically i'll keep calling their API until we have the entire list of assets and optimize for true pagination down the road.

> 2. Treat compressed and uncompressed as the same in the UI. I'm not married to this and would be interested in arguments against this.

i'm onboard with this as well. the only semantic differences between the two kinds of NFTs come on the backend when you're doing transfers (compressed NFTs require a proof from the API), but other than that the rest of the UI elements should be the same

jnwng:
if the code structure looks roughly okay, the next step would be to verify the end-to-end behavior once the Read API is returning the correct information—right now it's missing the off-chain metadata as well as some pagination elements

regarding feature flags, this is currently going to be a compile-time flag which is suitable for the short-term

Gajesh2007:
any help needed here?

jnwng:
i've revived this code and prepping it for an initial round of review. ETA later today , although it will not be a finished product

packages/common/src/solana/index.ts
```diff
@@ -298,6 +309,107 @@ export class Solana {
     });
   }
 
+  public static async transferCompressedNft(
```
jnwng, line 312: this is a pretty beefy transfer function that i expect to get moved to `@metaplex-foundation/js` and cleaned up. right now, though, this is the best i got.

packages/common/src/solana/programs/token.ts
```diff
@@ -89,12 +89,14 @@ export async function customSplTokenAccounts(
...
-      tokenMetadata,
-      new Map(mintsMap) as Map<string, RawMint>
-    );
+  const { fungibleTokens, fungibleTokenMetadata } = splitOutNfts(
+    tokenAccountsArray,
+    tokenMetadata,
+    new Map(mintsMap) as Map<string, RawMint>
+  );
+
+  // TODO(jon): Add a feature-flag here
```
jnwng, line 98: will probably use a feature flag to switch between the splTokenAccounts mode and this one

packages/common/src/solana/programs/token.ts
```diff
@@ -194,6 +196,115 @@ export async function fetchSplMetadata(
...
+      jsonrpc: "2.0",
+      method: "get_assets_by_owner",
+      // TODO(jon): This should uniquely identify this operation
+      id: "rpd-op-123",
+      params: [
+        // publicKey,
+        new PublicKey("6kPuZ1ZU684dwUSnJFXXFsCJfxKSaWnHRn2V5k4PgYm6"),
+        { sortBy: "updated", sortDirection: "desc" },
+        // change limit
+        5,
```
jnwng, line 213: TODO: will be automatically paginating through and retrieving _all_ NFTs

packages/common/src/solana/programs/token.ts
```diff
@@ -194,6 +196,115 @@ export async function fetchSplMetadata(
...
+      key: await getAssociatedTokenAddress(
+        new PublicKey(nft.id),
+        new PublicKey(nft.ownership.owner)
+      ),
+      compression: {
+        ...nft.compression,
+        ownership: {
+          ...nft.ownership,
+        },
+      },
```
jnwng, line 247: im adding this to the nftTokens object since that is what shows up on the other side of the "send" tokens flow, the metadata does not. i need this information to construct the transfer function, but if its better to just store this somewhere and retrieve it later, let me know

jnwng:
@Gajesh2007 @armaniferrante ready for another pass. mostly want to get feedback on where all the code is right now and whether there are better places for the code, the subsequent types, etc

work im starting now:
- [ ] paginating through all NFTs (only retrieving 5 right now)
- [ ] trying to get a minimal amount of the transfer logic into another SDK so Backpack doesn't have to have so much. what's here should work, provided the RPC returns the correct information!
- [ ] add a remote feature flag / feature detection on the connection to see if it responds to Read API methods
- [ ] will flag this entire feature on Developer Mode

PR #1312, opened 2022-10-28
Wallet Standard v1.0.0

jordansexton:
We've finally released the stable version. This PR updates Backpack to it. Expect semver updates only from here on out, and expect no more breaking changes for a while!


jordansexton:
Hmm, seems like CI died on an unrelated package?

armaniferrante:
> Hmm, seems like CI died on an unrelated package?

Probably unrelated. Reran.

PR #1325, opened 2022-10-28
cardinal creator royalties standard support

Gajesh2007:


packages/common/src/solana/index.ts
```diff
@@ -181,6 +182,90 @@ export class Solana {
...
+        createAssociatedTokenAccountInstruction(
+          walletPublicKey,
+          destinationAta,
+          destination,
+          mint
+        )
+      );
+    }
+
+    const tx = await withSend(
```
Gajesh2007, line 246: signature verification failed error

packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Solana/index.tsx
```diff
@@ -65,6 +66,13 @@ export function SendSolanaConfirmationCard({
           destination: new PublicKey(destinationAddress),
           amount: amount.toNumber(),
         });
+      } else if (await isCardinalWrappedToken(solanaCtx.connection, (token.mint?.toString() as string))) {
```
armaniferrante, line 69: We should do this fetch on component mount to speed things up. Can even add it to recoil as a hook + selector.

PR #1298, opened 2022-10-27
add pubkey back into slack notifier

wentokay:
\+ change casing of `publicKey` to `publickey` in auth worker for consistency

PR #1332, opened 2022-10-29
Stripe via popups

hkirat:
The PR adds three things 
1. Opens stripe in a popup whenever a user wants to unramp. Before this we showed it in the app-extension but that won't work until stripe supports iframes
2. Adds backend logic to store all stripe transactions, and listens to webhooks from stripe to store transaction status into the DB. This was originally written to show a better feedback to the user on the extension. We're not doing that anymore. This can be used in the future for users to be able to track their transactions right from their wallet.
3. Moves all `onramp` logic to the `auth` folder after discussion with the team as to the best place for it to be for now.


backend/workers/auth/src/zeus/const.ts
```diff
@@ -17,6 +25,7 @@ export const AllTypesProps: Record<string, any> = {
     _or: "auth_publickeys_history_bool_exp",
     blockchain: "String_comparison_exp",
     publickey: "String_comparison_exp",
+    user_id: "uuid_comparison_exp",
```
hkirat, line 28: cc @wentokay @tomlinton these seem unrelated to my changes. Could these be lingering from before?

wentokay: checking

wentokay:
I think the hasura change summary task failed to run because the github action is run in the context of your fork

so the `GITHUB_TOKEN` env var it uses isn't valid for https://github.com/coral-xyz/backpack

will look into this

backend/workers/auth/src/zeus/const.ts
wentokay: I just checked out this branch

removed the docker container `docker compose down -v`
restarted it `docker compose up`
waited for it to start up and ran `yarn zeus` inside the workers/auth dir and got these changes https://github.com/coral-xyz/backpack/commit/9bdd39c87724f935559d5edffa0060ebee79685d

wentokay: I could open it as a PR if you want to merge it into this branch?

hkirat: @wentokay yes that would be great, wonder how this got missed 🤔

hkirat:
@wentokay can I merge in this one? Would be good to get this in before the next release since we might need to make `stripe` public by then. 
If the DB migration stuff is causing issues maybe we can just run the migration script locally?

wentokay:
could you merge this commit into this branch please?

https://github.com/coral-xyz/backpack/commit/7ec9e69bcbae37e5c3661ba736221167ac781ce1 (branch: wen/stripe-amends)

it just runs `yarn` from root (`stripe` was missing from the root `yarn.lock`) and `yarn zeus` again from `backend/workers/auth`

backend/reef/hasura/migrations/default/1667001299142_create_table_auth_stripe_onramp/up.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE "auth"."stripe_onramp" ("id" serial NOT NULL, "publicKey" text NOT NULL, "client_secret" text NOT NULL, "webhook_dump" text NOT NULL, "status" text NOT NULL, PRIMARY KEY ("client_secret") );
```
wentokay, line 1: this is a minor thing and feel free to ignore for now but at some point could we make `stripe_onramp.publicKey` either `public_key` or `publickey`?

Ideally we keep to `snake_case` in the db just for consistency

The casing with `publickey/publicKey` is already [a bit messy](https://github.com/coral-xyz/backpack/pull/1298#discussion_r1006792140) so we can come back to this in future if we want to get this in ASAP

backend/workers/auth/src/zeus/const.ts
wentokay: pushed a new commit 7ec9e69bcbae37e5c3661ba736221167ac781ce1 to replace the previous one, but I can't seem to open a PR to be merged into this one as it's from your backpack fork and it doesn't seem to be possible in the github UI to do it 🤔

wentokay:
figured out how to make a PR here https://github.com/hkirat/backpack/pull/3

but I think you might've done the changes now anyway

Issue #1383, opened 2022-11-01
use token metadata standard to determine if something is an NFT


armaniferrante:
Currently we use token list and the existence of a metadata + uri to determine if something is an NFT or regular token. Instead we should use the token standard. 

Original request here: https://discord.com/channels/985994296337498182/1011846320677462137/1037126947156795452

armaniferrante:
Labeling as a P2. Not because it's not important, but because we are, in the next week, prioritizing things that are impairing users from using Backpack (e.g. blank screens, crashes, etc). We should do this soon once we work through those bugs.

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1828

PR #1388, opened 2022-11-02
Add support for metaplex fungible token metadata

ThornM9:
Referencing issue #1383


packages/recoil/src/atoms/solana/token.tsx
```diff
@@ -106,20 +111,61 @@ export const solanaTokenNativeBalance = selectorFamily<
     (tokenAddress: string) =>
     ({ get }: any) => {
       const tokenAccount = get(solanaTokenAccountsMap({ tokenAddress }));
-      if (!tokenAccount) {
```
armaniferrante, line 109: Why was this removed? Seems like it could be necessary still.

packages/recoil/src/atoms/solana/token.tsx
```diff
@@ -106,20 +111,61 @@ export const solanaTokenNativeBalance = selectorFamily<
...
+      const tokenMetadata: TokenMetadata | null = splTokenMetadata.find(
+        (m: TokenMetadataString) =>
+          metadataAddress.equals(new PublicKey(m.publicKey))
+      );
+      if (tokenMetadata) {
+        const displayBalance = ethers.utils.formatUnits(
+          nativeBalance,
+          tokenMetadata.decimals
+        );
+        const { symbol: ticker, image: logo, name } = tokenMetadata.uriMetadata;
```
tomlinton, line 137: This is not guaranteed to be defined I think?

ThornM9: True, updated that

packages/recoil/src/atoms/solana/token.tsx
ThornM9: Fixed

armaniferrante:
this has regressed the token list 

Screen Shot 2022-11-13 at 8 40 19 AM

ThornM9:
@armaniferrante that's fixed now, didn't realise some token registry tokens had a empty metaplex metadata account. Behaviour now is to only use metadata if it's defined and token registry is empty.

ThornM9:
@armaniferrante fixed, didn't realise that some token registry tokens would also have a metaplex metadata account. Behaviour is now to only use metaplex metadata if there's no token registry.

packages/common/src/solana/programs/token.ts
```diff
@@ -105,14 +131,92 @@ export async function customSplTokenAccounts(
...
+  const metadataArr = Array.from(tokens.values());
+
+  // if token standard is available then use it. otherwise determine fungibility by decimals
+  const tokensWithMetadata = metadataArr
+    .map((m) => {
+      return { ...m, decimals: m.decimals || 0 };
+    })
+    .filter((m) =>
+      m.metadata.tokenStandard
+        ? m.metadata.tokenStandard == TokenStandard.FungibleAsset ||
```
tomlinton, line 189: I think FungibleAsset should end up in the NFT view rather than grouped with SPL tokens, at least that is the guidance here https://docs.metaplex.com/programs/token-metadata/token-standard#introduction

packages/common/src/solana/programs/token.ts
```diff
@@ -73,26 +78,47 @@ export async function customSplTokenAccounts(
...
+    tokenAccountsArray,
+    tokenAccountToMint
   );
 
   //
-  // Fetch the metadata uri and interpert as NFTs.
+  // Fetch all metadata uris
   //
-  const nftMetadata = await fetchSplMetadataUri(
+  const tokenAccountsToMetadataWithUri = await fetchSplMetadataUri(
```
tomlinton, line 99: Tokens that metadata could not be fetched for are effectively filtered out of this array from the `fetchSplMetadataUri` function, that means they aren't separated into NFTs below. This is mostly spam NFTs, but we should try and separate them into NFTs correctly anyway. This will stop them showing up in the token list and provide the right access to burning options in the NFT detail pages.

Screenshot 2022-11-15 at 1 53 03 PM

tomlinton:
Left a few more comments @ThornM9, appreciate your effort here. There is a lot of edge cases with spammy tokens.

tomlinton:
This closes https://github.com/coral-xyz/backpack/issues/1352

tomlinton:
@ThornM9 will you be finishing this up? Happy to complete it for you if you are busy, but would like to get it in soon given some refactoring thats going to happen shortly.

ThornM9:
@tomlinton Really appreciate the feedback, would like to finish it off but been pretty busy. If it's holding you up, might be better if you take it up. Otherwise I'll probably be able to get to it early next week.

armaniferrante:
Completed here https://github.com/coral-xyz/backpack/pull/1828. Thank you for this! Was very helpful as a starting point!

PR #1409, opened 2022-11-02
chore: added types for element styles

anoushk1234:


packages/react-xnft/package.json
```diff
@@ -13,8 +13,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@coral-xyz/common-public": "*",
-    "@coral-xyz/themes": "*",
+    "@coral-xyz/common-public": "^0.2.0-latest.379",
```
hkirat, line 16: Could you revert this

yarn.lock
```diff
@@ -1634,7 +1634,24 @@
     stream-browserify "^3.0.0"
     util "^0.12.4"
 
-"@coral-xyz/xnft-cli@*", "@coral-xyz/xnft-cli@^0.0.1":
+"@coral-xyz/common-public@^0.2.0-latest.379":
```
hkirat, line 1637: Could you revert this?

hkirat (changes requested):

yarn.lock
anoushk1234: sure, although it wasn't allow me to build without these

anoushk1234:
@hkirat is this gtg?

packages/react-xnft/package.json
```diff
@@ -13,8 +13,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@coral-xyz/common-public": "*",
```
hkirat, line 16: I think you ended up deleting these by mistake? Can you add those back

yarn.lock
hkirat: I still see  the yarn changes, could you revert to the current lock file?

hkirat (changes requested):

packages/react-xnft/src/reconciler.ts
```diff
@@ -710,7 +710,23 @@ export const HOST: Host = {
   })(),
 };
 
-type Style = any;
+type _HostConfig = HostConfig<
```
hkirat, line 713: I think this flowed over from a merge conflict, mind removing this as well? It's dead code

anoushk1234: yea i think this is from your recent pr, i synced this before pushing and fixed the conflicts as well, ill remove it

PR #1416, opened 2022-11-02
Reconciler cleanup

hkirat:
The PR cleans up the reconciler. 
We now pass down all props that the user gives us to the renderer. This isn't a problem anymore since all this code runs inside an IFRAME. The renderer can chose to use/throw the props it doesn't need.

This also ensures that we propagate updates for all props to the reconciler. Until now we were cherrypicking the props whose updates we want to propagate based on the NodeKind (onClick, style ...)

### DIFF
--- a/packages/react-xnft/src/ReactDom.ts
+++ b/packages/react-xnft/src/ReactDom.ts
@@ -115,51 +115,8 @@ export class ReactDom {
       throw new Error("element not found");
     }
 
-    switch (instance.kind) {
-      case NodeKind.View:
-        if (updatePayload.style) {
-          instance.style = updatePayload.style;
-        }
-        break;
-      case NodeKind.Text:
-        if (updatePayload.style) {
-          instance.style = updatePayload.style;
-        }
-        break;
-      case NodeKind.TextField:
-        if (updatePayload.value !== undefined && updatePayload.value !== null) {
-          instance.props.value = updatePayload.value;
-        }
-        break;
-      case NodeKind.NavAnimation:
-        if (
-          updatePayload.routeName !== undefined &&
-          updatePayload.routeName !== null
-        ) {
-          instance.props.routeName = updatePayload.routeName;
-        }
-        break;
-      case NodeKind.Path:
-        if (updatePayload.fill !== undefined && updatePayload.fill !== null) {
-          instance.props.fill = updatePayload.fill;
-        }
-        break;
-      case NodeKind.Button:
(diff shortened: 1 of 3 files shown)


packages/react-xnft/src/reconciler.ts
```diff
@@ -925,21 +710,6 @@ export const HOST: Host = {
   })(),
 };
 
-type _HostConfig = HostConfig<
```
hkirat, line 928: dead code

packages/react-xnft/src/reconciler.ts
```diff
@@ -1279,15 +1058,13 @@ type DefNodeSerialized<K, P> = {
...
   props: P;
-  style: Style;
   children: Array<Element>;
 };
 
 type CustomNodeSerialized = {
   id: number;
   kind: NodeKind.Custom;
   props: CustomProps;
-  style: Style;
```
hkirat, line 1290: style is now part of the props

packages/react-xnft/src/ReactDom.ts
```diff
@@ -115,51 +115,8 @@ export class ReactDom {
...
-          instance.style = updatePayload.style;
-        }
-        if (updatePayload.src) {
-          instance.props.src = updatePayload.src;
-        }
-        break;
-      default:
-        throw new Error("invariant violation");
-    }
+    instance.props = { ...updatePayload.props };
```
armaniferrante, line 118: Does the style get properly propagated? The deleted code here explicitly sets the style field and that is no longer set.

hkirat: Yeah styles are basically part of `props` now.
This is how I'm extracting styles now for the components being rendered - https://github.com/coral-xyz/backpack/pull/1416/files#diff-60ff5434f0c7e93bc972f52821f707c3e9e5995271cfe9a82c2c1b455e1dd39bR161

hkirat: Just confirmed locally that they do get updated with the following code - 
```js
  const [color, setColor] = useState("red");

  useEffect(() => {
    window.setInterval(() => {
      setColor(c => c === "red" ? "yellow" : "red")
    }, 1000);
  }, []);

return <Text style={{color: color}}>hi there</Text>

```

tomlinton:
Screen Shot 2022-11-02 at 10 00 44 PM 

My kind of PR 🚀

Issue #1441, opened 2022-11-08
react-xnft: file input


Gajesh2007:
https://github.com/coral-xyz/xnft-wishlist/issues/1

jim4067:
this is a great idea. I noticed that you cannot import files such as fonts and images at the moment. 

Any updates after closing #1443?

armaniferrante:
Closing as we're moving to react native.

PR #1443, opened 2022-11-08
added FileInput

Gajesh2007:
#1441 

- in progress


Gajesh2007:
yet to test

jokumusic:
I tried to add some updates to your PR, but I don't have permission to. I'm not sure of the right way to do that. 
I added some changes to your PR to get it exported and usable by apps.
I was unable to test, because I think the backpack extension needs to be built with the changes, so that they can be tested. I wasn't able to successfully build the backpack extension to test.

hkirat:
@Gajesh2007 can you merge in the commits from https://github.com/coral-xyz/backpack/pull/1470 in here?

armaniferrante:
How does react native handle file input? Should this be a custom component instead?

jokumusic:
> How does react native handle file input? Should this be a custom component instead?

I was thinking that any of the various react native filepicker components can be rendered. It would be the same for things like drop downs, checkboxes, calendars, etc...

Issue #1359, opened 2022-10-31
add popout button in the xnft toolbar menu


wentokay:
as a user i want to be able to launch backpack into a popup window while I'm using an xNFT

PR #1452, opened 2022-11-10
Expose fullscreen functionality to xnfts

hkirat:
Useful for apps that need more real estate than just the popup width and height.

TODO: Add a parameter here that lets developers pop out video (small size, but popped out)


hkirat:
Related https://github.com/coral-xyz/backpack/issues/1359

PR #1453, opened 2022-11-10
useDimensions hook

jokumusic:
listens to window resize events and returns {width: window.innerWidth, height: window.innerHeight}


hkirat:
Thankyou for the PR @jokumusic this is great.
TODO: Add mobile implementation as well/ensure this works there.

PR #1457, opened 2022-11-11
change default solana rpc

wentokay:


packages/background/src/backend/core.ts
```diff
@@ -190,6 +190,25 @@ export class Backend {
...
+      await setWalletData({
+        ...data,
+        solana: {
+          ...data.solana,
+          cluster: SolanaCluster.DEFAULT,
+        },
+      });
+    }
+
     return (data.solana && data.solana.cluster) ?? SolanaCluster.DEFAULT;
```
armaniferrante, line 212: This will still return the old URL, even if igrated right?

wentokay: I don't think so? but i'm probably being stupid.

`SolanaCluster.DEFAULT !== OLD_DEFAULT` probably isn't necessary, I only added it because I thought it might be needed during the transition if people are switching branches in development.

I think I'll revert the [last commit](https://github.com/coral-xyz/backpack/pull/1457/commits/1a3b145adbf7ef548411198c1ca622705f830257)

armaniferrante: `data` never gets set on the migration. So it will always be the old URL. We want to return the new, migrated URL.

wentokay: it worked for me in testing because i closed and reopened the extension, so the method was called a second time

fixing now

PR #1459, opened 2022-11-11
solana connection serialization fix between service worker / mobile app

peterpme:
fixes the solana serialization between mobile app and service worker, originally implemented by armani


armaniferrante:
Will immediately merge this after cutting the next release.

PR #1478, opened 2022-11-13
add click/key listeners to update last used ts to fix autolock

tomlinton:


packages/recoil/src/hooks/client.tsx
```diff
@@ -28,3 +29,22 @@ export function useBackgroundKeepAlive() {
...
+export function useUpdateLastUsed() {
+  const background = useBackgroundClient();
+  const updateLastUsed = () => {
+    background.request({
+      method: UI_RPC_METHOD_KEYRING_UPDATE_LAST_USED,
+      params: [],
+    });
+  };
+
+  useEffect(() => {
```
tomlinton, line 42: @armaniferrante I don't really like this. I thought about hooking navigation state changes which would work, except for if the user has an xnft open. Do you remember anything about how the autolock implementation was intended to work?

armaniferrante: Instead of adding an RPC API, why not just update the last used from the relevant RPC endpoints? For example, if you call any of the UI_RPC_METHOD_* apis (though check me on this), then the autolock should be reset.

tomlinton: That was my first approach, but then I realised there may be no RPC calls if you have an xnft open. This click/keypress listener also doesn't register for things like flappy bad unfortunately.

The best approach might be to reset autolock on UI_RPC_METHOD calls (except things like polling updates) and disable autolock if an xnft is open.

PR #1501, opened 2022-11-15
re-introduce expo sdk 47 updates with react 18

peterpme:
reintroduces expo 47 w react 18 (working on my end)


wentokay:
i don't think xNFTs are working in this branch yet due to the different react version conflicts

Screenshot 2022-11-15 at 12 24 11

PR #1508, opened 2022-11-15
Adds push notifications to app-extension

hkirat:
The PR adds a way for us to push notifications to users. 
It also sets the foundation for xnft developers to send push notifications to users if the user has allowed the same for an xnft
Screenshot 2022-11-15 at 8 32 58 PM

This can be useful for pushing on chain events like 
1. Someone just bought your nft
2. Someone just bid on your nft
3. Price alerts?

Few xnft ideas I can think of are
1. Whale alert xnft
2. Peer to peer calls (when someone calls you, a notification pops up clicking on which connects you to the other user)
3. Subscribe to notifications from trading experts

We need to get JWTs for this to become secure


hkirat:
Plan to merge this in soon (hopefully today). No changes on the UI/app-extension yet so it should be safe even if we're cutting a release after merging this in. 
Implementation is insecure but will become secure as soon as JWT is in. Infra has already been deployed on fly.

PR #1513, opened 2022-11-16
monorepo eslint config

tomlinton:


tomlinton:
@johnrees this seems to work pretty well as a config. `yarn run lint:fix` from root to see it do its magic.

PR #1518, opened 2022-11-17
run yarn in root

wentokay:
+ remove an unused recoil dep


wentokay:
i think we need to run `yarn install` from root to get the lockfile up to date but there's something wrong with my `@coral-xyz/xnft-cli` versioning locally and it's downgrading the version

someone else might need to do it

Screenshot 2022-11-17 at 00 42 06

PR #1537, opened 2022-11-18
implemented the rest of the Solana Connection interface methods.

jokumusic:


packages/background/src/backend/solana-connection.ts
```diff
@@ -608,293 +608,337 @@ export class SolanaConnectionBackend {
     return await this.connection!.getAddressLookupTable(programId, config);
   }
 
-  ///////////////////////////////////////////////////////////////////////////////
```
armaniferrante, line 611: Note that this is only partially implemented and won't work as is. You still need to implement the `/frontend/` handlers and the connection client in `/packages/common/`

jokumusic: I purposely didn't do the frontend implementation because I wasn't sure if there was a design decision to not expose some methods. For instance, you might not want the subscribe methods exposed which could cause extra load on your RPC servers when people don't tear them down when not in use.

jokumusic: If you give the word, I'll go ahead and expose all the methods to the front end :)

armaniferrante: Let's do everything except for subscriptions and anything requiring websockets.

packages/background/src/frontend/solana-connection.ts
```diff
@@ -64,6 +112,7 @@ import * as bs58 from "bs58";
 import { decode } from "bs58";
 
 const logger = getLogger("solana-connection");
+const messageHandlers: Map<string, Function> = new Map();
```
armaniferrante, line 113: This is nice but it introduces an inconsistency not found in any of the other APIs. Personally I would recommend removing this but also will let others chime in.

jokumusic: I noticed a few of the frontend methods were passing what looks like the wrong parameter index, but maybe I'm misunderstanding something. For example, look at getLatestBlockhash() in frontend/solana-connection.ts and solana/background-connection.ts in common:
image

Shouldn't the frontend call look like ctx.backend.getLatestBlockhashAndContext(params[0]) instead?

armaniferrante: yes that sounds right

jokumusic: I've done the grunt work. I don't have any tests that cover all of the methods. Is there a test somewhere that covered the previously existing methods? With most of the methods, I just passed the channel message params straight through to the ackend connection because I didn't know if any serialization/deserialization or something else was needed. Let me know if there's more that I need to do.

packages/background/src/frontend/solana-connection.ts
```diff
@@ -478,3 +666,426 @@ async function handleGetAddressLookupTable(
...
+  ctx: Context<SolanaConnectionBackend>,
+  to: string,
+  lamports: number
+) {
+  const resp = await ctx.backend.requestAirdrop(new PublicKey(to), lamports);
+  return [resp];
+}
+
+/*
+async function handleSendTransaction(
```
hkirat, line 1070: could u remove the comment ehre

packages/background/src/frontend/solana-connection.ts
```diff
@@ -277,17 +464,21 @@ async function handleConfirmTransaction(
   return [resp];
 }
 
-async function handleGetMultipleAccountsInfo(
+/*
+async function handleSimulateTransaction(
```
hkirat, line 468: can u remove the commend here?

hkirat (changes requested):

packages/background/src/frontend/solana-connection.ts
jokumusic: I commented it out because the arguments to the function aren't primitive types and probably need to be deserialized. I'm not really sure how to go about that right now.

packages/background/src/frontend/solana-connection.ts
jokumusic: I commented it out because the arguments to the function aren't primitive types and probably need to be deserialized. I'm not really sure how to go about that right now.

packages/background/src/frontend/solana-connection.ts
jokumusic: Also, the Signer type contains public/private key, so I don't think the signature of the function is correct. It should probably just be the pubkey managed by the wallet and then the wallet will confirm the signature.

packages/background/src/frontend/solana-connection.ts
jokumusic: Also, the Signer type contains public/private key, so I don't think the signature of the function is correct. It should probably just be the pubkey managed by the wallet and then the wallet will confirm the signature.

hkirat (approved):

PR #1540, opened 2022-11-18
Added the chat sdk and backend

hkirat:
The PR adds
1. A new databse
2. A new hasura instance
3. A client side sdk that can be used by `app-extension`/ other xnfts to add realtime chat features.


wentokay (approved):
lgtm!

some questions but none are these are blockers if you want to merge right away!

- do you think `rooms` should be a separate table or add an index on the `room` field? I assume that's going to be the address of an xnft?
- there's a serial ID and a uuid, is this to take advantage of incremental ID ordering, or is the uuid the user's id? if it's the ordering thing, it might be worth adding a `created_at` timestamp on chat messages at some point? And maybe switch to uuid as the primary key in case we need to run other infra? It'd also mean we could show when messages were posted but maybe I'm missing something?

future stuff that might be useful, or not
- a length constraint on the `chats.message` field
- an enum for chats.status in future with something like `visible` | `hidden` to start with

hkirat:
@wentokay super cool suggestions I think I can add all of these in the same PR.
The `uuid` is the users `uuid` that is correct. 
The incremental integer id is so that when we receive graphql subscription updates, we can merge in the messages correctly as they come in bulk from the subscription.
`created_at` is definitely a field that should be there irrespective of that.

wentokay:
>`created_at` is definitely a field that should be there irrespective of that.

sweet, there's a helper in the hasura console when adding a created_at field, it's under suggested items or something

PR #1579, opened 2022-11-24
add JWT auth

wentokay:
> **Note**
> if you are testing this locally, before running you will need to -
> - set `BACKPACK_FEATURE_JWT` and `BACKPACK_FEATURE_USERNAMES` to `true`
> - find and replace `https://auth.xnfts.dev/` to `http://localhost:8787/`
> - `cd backend/workers/auth && yarn dev`

```mermaid
sequenceDiagram
    Note over User (Backpack UI): 🔒 Unlock Screen
    User (Backpack UI)->>Backpack:User submits password to unlock
    Backpack->>+Auth Worker: POST /authenticate/:username (empty body)
    alt no JWT cookie included with the request
        Note over Auth Worker: fetch { id, publickeys } from hasura for :username
        Auth Worker->>Backpack: return { id, publickeys }
        Note over Backpack: Sign a message using one of the provided { publickeys }
        Backpack->>+Auth Worker: POST signed message to /authenticate
        Note over Auth Worker: Verify signed message
        alt ✅ Signature is valid & username/id/publickey are OK
            Auth Worker->>Backpack: 200 OK (including new JWT cookie)
            Backpack->>+User (Backpack UI): User is logged in, with JWT stored in cookie
            Note over User (Backpack UI): 🔓 Balances Screen
        else ❌ Signature is invalid or issue with username/id/publickey
            Auth Worker->>Backpack: 400 Error
            Backpack->>+User (Backpack UI): Display error
            Note over User (Backpack UI): 🔒 Unlock Screen (+ Error... TODO)
        end
    else ✅ request included a valid JWT cookie
        Auth Worker->>Backpack: 200 OK
        Backpack->>+User (Backpack UI): User is logged in, with JWT stored in cookie
        Note over User (Backpack UI): 🔓 Balances Screen
    else ❌ request included an invalid JWT cookie
        Auth Worker->>Backpack: 401 Unauthorized (+ remove JWT cookie)
        Backpack->>+User (Backpack UI): Display error, remove JWT from cookies
        Note over User (Backpack UI): 🔒 Unlock Screen (+ Error... TODO)
    end
```

TODO:
- [x] return jwt on a new signup
- [x] clear JWT after resetting backpack
- [x] move the client signin flow somewhere else, currently breaks if user enters password incorrectly initially
- [x] prevent signin if the cookie is invalid (related to first point)
- [x] decide if we're ok with backpack only being unlockable with a working internet connection?
- [ ] wait for signature from ledger (related to two points)
- [ ] add a refresh token for invalidating JWT session
- [ ] add some UI to instruct ledger users what's going on
- [ ] display error in UI when server rejects invalid cookie
- [ ] prevent user from removing the pubkey we are associating with their username in the db
- [ ] add a unique audience scope to JWT for xNFTs
- [ ] consider other edge cases?


packages/app-extension/src/app/Router.tsx
```diff
@@ -314,16 +317,65 @@ function WithEnabledBlockchain({
...
+        });
+      }
+    };
+
+    if (BACKPACK_FEATURE_JWT) {
+      ensureUserHasJWT().then(() => setIsAuthenticated(true));
+    } else {
+      setIsAuthenticated(true);
+    }
+  }, [isLocked]);
```
wentokay, line 372: this will be moved elsewhere, it was convenient for quick testing

armaniferrante:
>  decide if we're ok with backpack only being unlockable with a working internet connection?

This should only be the case for a) the initial migration and b) for onboarding. I think for both this is fine. In the more general sense, backpack should be unlockable without an internet connection.

armaniferrante:
> ok I might be missing something obvious but this might mean we'd have to do client-side JWT validation, so it couldn't be stored in a httponly cookie

Yea you're right. Let's just require the internet then for v0.

backend/workers/auth/src/index.ts
```diff
@@ -409,3 +416,141 @@ const camelCaseToSentenceCase = (str: string) =>
...
+        return c.json({ msg: "valid jwt cookie" }, 200);
+      } else {
+        throw new Error(`invalid username (${username})`);
+      }
+    } catch (err) {
+      console.error(err);
+      clearCookie(c, "jwt");
+      return c.json({ msg: "invalid jwt cookie" }, 401);
+    }
+  } else {
```
hkirat, line 536: Could you explain the reason for the `else` here?

wentokay: if it wasn't included and someone had a JWT in stored in a cookie for user `a`, and they called `/authenticate/b`, the server would return `{ msg: "valid jwt cookie" }`

if you don't think that check is necessary I'm happy to take it out though

wentokay: oh wait, you're referring to a different `else`, looking now

wentokay: ok so this `else` block is invoked when a user first calls `/authenticate/:username` (most likely an existing user, logging in for the first time after receiving this update) so they don't have a JWT cookie yet

the server responds with the `id` and `publickeys` for that username, expecting the client to make one more call to `POST /authenticate` including a signed message from one of the `publickeys`

if they do this correctly, the next time they log in they will have a JWT cookie attached to the request, so the if block will be called instead of the else block

packages/app-extension/src/app/Router.tsx
```diff
@@ -314,16 +317,65 @@ function WithEnabledBlockchain({
...
+          },
+        }
+      );
+      if (res.status !== 200) throw new Error("failed to authenticate");
+
+      const body = await res.json();
+      if (body.id && body.publickeys?.length) {
+        const signatureBundle = await background.request({
+          method: UI_RPC_METHOD_TRY_TO_SIGN_MESSAGE,
+          params: [
```
hkirat, line 349: This requests automatically signs a transaction for a user for `any arbit message` (without a user prompt) ?
Could that be a security concern (I could sign a message that says `login to xyz.com at timestamp T` without the user knowing)?

armaniferrante: Resolved offline.

hkirat:
@wentokay is this okay to merge as is? Or would you be merging it after the rest of the `checkbox` tasks are done? 
For my usecases (chat and notifications) this seems to be enough (just tested it by deploying to cloudflare as well) .

wentokay:
> @wentokay is this okay to merge as is? Or would you be merging it after the rest of the `checkbox` tasks are done? For my usecases (chat and notifications) this seems to be enough (just tested it by deploying to cloudflare as well) .

I'm getting very inconsistent results with testing the JWT setup with a ledger at the moment. I can't get message signing to work during the onboarding flow either so it might be related to my setup.

I could do a couple of small tweaks in a commit and merge this with `BACKPACK_FEATURE_JWT` disabled by default, then you could enable it when you're doing your work if that helps?

I could work on the remaining checkboxes in a follow-up PR

wentokay:
Merging for now behind a feature flag (`BACKPACK_FEATURE_JWT`) to unblock development on other branches. The remaining checkbox items will be addressed in a separate PR

PR #1583, opened 2022-11-24
fix: catch API request fails

### LINKED ISSUE #1009: App crashes if `coingecko` requests fail
I think coingecko is down in India right now and I'm seeing this consistently. 
If I try to send a transaction (lets say using the `provider-test` example), the app crashes. 
The issue is that coingecko returns a 502 with HTML response and we don't catch any of that [here](https://github.com/coral-xyz/backpack/blob/master/packages/recoil/src/atoms/prices.tsx#L166)

Screenshot 2022-10-04 at 11 25 50 PM
Screenshot 2022-10-04 at 11 25 59 PM

This can probably be handled more gracefully

jim4067:
resolves #1009

### DIFF
--- a/packages/recoil/src/atoms/prices.tsx
+++ b/packages/recoil/src/atoms/prices.tsx
@@ -160,7 +160,7 @@ const pricesForIds = selector({
       );
     } catch (err) {
       console.error("error querying all Coingecko IDs", err);
-      return;
+      return new Map();
     }
   },
 });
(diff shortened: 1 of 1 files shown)


packages/recoil/src/atoms/prices.tsx
```diff
@@ -145,15 +145,23 @@ const pricesForIds = selector({
...
+      return new Map(
+        // Transform the response from id -> price data to addresss -> price data
+        Object.keys(json).map((id) => [
+          coingeckoIdToAddressMap.get(id),
+          json[id],
+        ])
+      );
+    } catch (err) {
+      console.error("error querying all Coingecko IDs", err);
+      return;
```
armaniferrante, line 163: Should this return `new Map()` instead?

jim4067: It should. Let me fix that.

Issue #1599, opened 2022-11-27
update username associations everytime a pubkey is created or deleted


armaniferrante:
P1 because this will cause a bunch of edge cases if we don't do this soon.

armaniferrante:
Note: to do this we will need a migration step for all imported keys to register it to the username.

armaniferrante:
Note: we also need to add a "sign message" drawer to the ledger import flow as well (in the same way we do during onboarding).

PR #1604, opened 2022-11-28
check derived pubkeys when signing JWT

wentokay:
problem:

a user might have removed the pubkey we store/associate with their username when they created their account

solution:

when locally signing a message to acquire a JWT, this checks against 20 derived pubkeys under each derivation path

the code could be improved a bit, including only checking the derived pubkeys if there is no match in the active wallet addresses, but given that the function is only called once and that #1599 is in the works, I figured it might be good enough for now

### DIFF
--- a/packages/background/src/backend/core.ts
+++ b/packages/background/src/backend/core.ts
@@ -433,11 +433,13 @@ export class Backend {
   /**
    * attempts to sign a message with one of public keys provided, it will try to use
    * a non-ledger pubkey if that's a possibility
+   * @param password password to unlock mnemonic for deriving pubkeys
    * @param msg message to be signed
    * @param publicKeysIncludingBlockchain Array<{ blockchain: string; publickey: string; }>
    * @returns
    */
   async tryToSignMessage(
+    password: string,
     msg: string,
     publicKeysIncludingBlockchain: Array<{
       blockchain: Blockchain;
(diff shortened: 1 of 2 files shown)


packages/background/src/backend/core.ts
```diff
@@ -446,35 +448,90 @@ export class Backend {
...
+
       // fetch all keys in the store
       const keys = await this.keyringStore.publicKeys();
 
+      // fetch 20 keys from each derivation path in case the user
+      // has deleted the pubkey that we associate with their username
+      const derivedKeys = (async () => {
+        try {
+          const blockchains = [Blockchain.SOLANA, Blockchain.ETHEREUM];
+          const derivPaths = Array.from(new Set(Object.values(DerivationPath)));
```
wentokay, line 460: this just ensures that the array contains unique values

wentokay:
strange side-effect of this PR that seems to be happening

- start with wallet `A`
- add wallet `B`
- remove wallet `A`
- lock backpack and remove JWT cookie
- log in (backpack will match and sign with wallet `A` in the background, by searching the derivation path)
- close wallet
- reopen backpack, wallet `A` has reappeared in the list

packages/background/src/backend/core.ts
```diff
@@ -433,11 +433,13 @@ export class Backend {
   /**
    * attempts to sign a message with one of public keys provided, it will try to use
    * a non-ledger pubkey if that's a possibility
+   * @param password password to unlock mnemonic for deriving pubkeys
    * @param msg message to be signed
    * @param publicKeysIncludingBlockchain Array<{ blockchain: string; publickey: string; }>
    * @returns
    */
   async tryToSignMessage(
```
armaniferrante, line 441: Can we make this private? It gives me some pause signing messages on a public endpoint with keys not in the keyring store.

armaniferrante:
Instead of searching for the key to sign, would it make more sense to just treat the user as logged out and force onboarding again?

wentokay:
closing this after a discussion about usernames and onboarding

PR #1621, opened 2022-11-29
ANS integration by Onsol

crypt0miester:
Enabled sending to ANS domains by Onsol.

currently, only works on mainnet.


hkirat:
Hi @crypt0miester , just wanted to let you know the status on this. Since the PR introduces a new dependency (`@onsol/tldparser`) we might have to audit the library since it runs in the same context as the `app-extension`

crypt0miester:
> Hi @crypt0miester , just wanted to let you know the status on this. Since the PR introduces a new dependency (`@onsol/tldparser`) we might have to audit the library since it runs in the same context as the `app-extension`

thank you for your reply.

here’s a link to the repo: https://github.com/onsol-labs/tld-parser

let me know if there’s anything else you need from us 🤝

crypt0miester:
really appreciate this. @armaniferrante

PR #1569, opened 2022-11-22
feat: Re-use @coral-xyz/zeus in workers/auth

jim4067:
Resolves #1554

PR #1623, opened 2022-11-29
fix zeus import path in auth worker

wentokay:
I think the zeus directory has been removed from the auth worker


wentokay:
related #1569

Issue #1632, opened 2022-11-30
merge auth worker into backpack-api


tomlinton:
https://github.com/coral-xyz/backpack/tree/master/backend/workers/auth into https://github.com/coral-xyz/backpack/tree/master/backend/native/backpack-api

tomlinton:
Closes by #1641

PR #1641, opened 2022-12-01
migrate jwt methods from auth worker to backpack-api

### LINKED ISSUE #1615: remove username from jwt
```
{
  sub: 'd9a23634-1cb9-4993-931c-729746e29035',
  username: 'god',
  iss: 'auth.xnfts.dev',
  aud: 'backpack',
  iat: 1669670538
}
```

tomlinton:
Ref #1632, makes sense to move these now because then I don't have to test both builds against the old worker.

Removes username from JWT, closes #1615.


backend/native/backpack-api/src/db/users.ts
```diff
@@ -9,53 +9,118 @@ const chain = Chain(HASURA_URL, {
...
+export const getUser = async (id: string) => {
+  const response = await chain("query")({
+    auth_users_by_pk: [
+      {
+        id,
+      },
+      {
+        id: true,
+        username: true,
+        public_keys: [{}, { blockchain: true, public_key: true }],
```
tomlinton, line 58: @hkirat i'm returning more data here

backend/native/backpack-api/src/db/users.ts
```diff
@@ -9,53 +9,122 @@ const chain = Chain(HASURA_URL, {
 
 export const getUsers = async (
   userIds: string[]
-): Promise<{ username: string; id: string }[]> => {
+): Promise<{ username: unknown; id: unknown }[]> => {
   const response = await chain("query")({
     auth_users: [
       {
-        where: { id: { _in: userIds } },
+        where: { username: { _in: userIds } },
```
hkirat, line 16: Could you explain this change? 
The places where I've used `getUsers` I want to fetch based on userIds and not the username

backend/native/backpack-api/src/routes/v1/authenticate.ts
```diff
@@ -0,0 +1,55 @@
...
+import {
+  validateSolanaSignature,
+  validateEthereumSignature,
+} from "../../validation/user";
+
+const { base58 } = ethers.utils;
+
+const router = express.Router();
+
+router.delete("/authenticate", async (req, res) => {
```
hkirat, line 14: Would this now be a `delete` to `backpack-api.xnfts.dev/authenticate/authenticate` ? 
Since in `index.ts` we're routing as follows - 
```
app.use("/authenticate", authenticateRouter);
```

backend/native/backpack-api/src/routes/v1/authenticate.ts
```diff
@@ -0,0 +1,55 @@
...
+  clearCookie(res, "jwt");
+  return res.json({ msg: "ok" });
+});
+
+router.post("/authenticate", async (req, res) => {
```
hkirat, line 19: Same as the comment above, this would now get triggered on `/authenticate/authenticate`

packages/app-extension/src/components/Onboarding/pages/RecoverAccountUsernameForm.tsx
```diff
@@ -22,17 +23,14 @@ export const RecoverAccountUsernameForm = ({
...
-          {
-            headers: {
-              "x-backpack-waitlist-id": getWaitlistId() || "",
-            },
-          }
-        );
-        const json = await res.json();
+        const response = await fetch(`${BACKEND_API_URL}/users/${username}`, {
+          headers: {
+            "x-backpack-waitlist-id": getWaitlistId() || "",
```
hkirat, line 29: Is `x-backpack-waitlist-id` being used at all now? Might be missing something but can't find it anywhere in the changes you've made

hkirat (changes requested):

backend/native/backpack-api/src/db/users.ts
tomlinton: Reverted thanks, had a big merge collision with that recent PR and I must have messed up

packages/app-extension/src/components/Onboarding/pages/RecoverAccountUsernameForm.tsx
tomlinton: I don't believe so. I didn't change this, it showed in the diff because of reformatting, but I might as well remove it since it doesn't look like its used anywhere in `master` either

tomlinton: cc @johnrees

PR #1655, opened 2022-12-02
Allow one_xnft role to read username from id

greenish:


armaniferrante:
can we merge this?

greenish:
Yes will merge and migrate today

PR #1667, opened 2022-12-02
Fix "plugin not found"

greenish:


packages/app-extension/src/components/Unlocked/Apps/Plugin.tsx
```diff
@@ -74,11 +63,13 @@ export function _PluginDisplay({
       }}
     >
       <PluginControl closePlugin={closePlugin} />
-      <PluginRenderer
-        key={plugin.iframeRootUrl}
-        plugin={plugin}
-        xnftPreference={xnftPreference}
-      />
+      {plugin && (
```
tomlinton, line 66: Is this the main fix here and the rest is fixing the typing from changing from our custom hook to a recoil hook?

greenish: I mostly needed the hook not return a default object so I can safely test if the plugin and preferences exists or not.

PR #1645, opened 2022-12-01
More chat follow ups

hkirat:
More chat stuff. 
Message requests, jwt auth in chat
Feature gates
Pagination of messages as u scroll up

PR #1695, opened 2022-12-04
Fixed DB schema

hkirat:
Bunch of inconsistencies were introduced in https://github.com/coral-xyz/backpack/pull/1645
1. https://github.com/coral-xyz/backpack/pull/1645/files#diff-76d32ad3a0460147e7f0fa5786866356fbc2f1eddd5ce789ed0083d9a2b89d6bL5 - No idea how/why this would this got changed but this is what broke the account creation flow
2. https://github.com/coral-xyz/backpack/pull/1645/files#diff-93765eb6c4a712b34e907a4b5e1eb608c585bbec6ed46b03e38eaf5f0701b842R8 -- This again no idea how it made it's way in there, seems like some way of tracking `public_keys` rather than it being an entirely new table

This affected prod for the last ~2 hours or so, was seeing this locally before that (prod was unaffected before that).


backend/reef/hasura/metadata/databases/default/tables/auth_users.yaml
```diff
@@ -1,15 +1,31 @@
...
+  - name: waolist_token
+    using:
+      foreign_key_constraint_on:
+        column: username
+        table:
+          name: waolist_tokens
+          schema: auth
 array_relationships:
-  - name: publickeys
+  - name: public_keys
```
hkirat, line 23: This change was the major culprit, introduced in https://github.com/coral-xyz/backpack/pull/1645
My guess is a global replace of this happened on my machine causing it to change here

tomlinton:
Possibly resulting from a local hasura instance that wasn't synced with the current metadata and migrated? `npx hasura metadata export` / `yarn run zeus` maybe, sometimes its hard to manage with both local and remote states

PR #1741, opened 2022-12-07
Backend for notifications

hkirat:
1. Make notifications secure
2. Provide xnft developers with a secret that they can get from the xnft.gg dashboard and use to talk to the `xnft-api-server`
3. Make all routes use auth and ensure secret is being passed.


backend/native/xnft-api-server/src/auth/middleware.ts
```diff
@@ -1,15 +1,53 @@
...
+  if (
+    timestamp < new Date().getTime() ||
+    timestamp > new Date().getTime() + 60 * 1000
+  ) {
+    return res.status(403).json({ msg: "Expired message, please try again" });
+  }
+  const message = `You are trying to log into Backpack at ${timestamp}`;
+
+  try {
+    const verified = nacl.sign.detached.verify(
```
hkirat, line 40: @armaniferrante @tomlinton thoughts on this? 
Basically I'm making the `publisher` of an xnft `sign a message` on `xnft.gg` to validate their identity on our backend and allowing them to create a `secret` (that we store in our DB) for all future requests governing that xnft (send notification to these xnft users being a request that they can send)

Issue #1761, opened 2022-12-08
can't remove wallets


tomlinton:
Try to delete any wallet via  `Your account` / `Edit wallets`:

<img width="748" alt="Screenshot 2022-12-08 at 3 56 05 PM" src="https://user-images.githubusercontent.com/489202/206345026-1dfe8254-f048-4f9f-a8f4-eabb6ce31de5.png">

Issue #1760, opened 2022-12-08
can't add new derived wallet or imported wallet


tomlinton:
Click create new wallet and observe crash

<img width="367" alt="Screenshot 2022-12-08 at 3 53 29 PM" src="https://user-images.githubusercontent.com/489202/206344649-22d06849-d7b5-4615-b535-2b60503fa037.png">

<img width="382" alt="Screenshot 2022-12-08 at 3 53 33 PM" src="https://user-images.githubusercontent.com/489202/206344654-e93fe730-1b59-4f53-82a5-3d03a23cfdf3.png">

tomlinton:
Also generates the same read only error as https://github.com/coral-xyz/backpack/issues/1761 so it is probably the same issue

PR #1763, opened 2022-12-08
fix wallet create and delete

tomlinton:
Use copies to fix read only errors

Closes https://github.com/coral-xyz/backpack/issues/1761
Closes https://github.com/coral-xyz/backpack/issues/1760


tomlinton:
Merging this @armaniferrante

Issue #1694, opened 2022-12-04
logout username


armaniferrante:
* for multiple accounts, logout should logout the existing account
* for the last account, logout should be equivalent to reset wallet

PR #1766, opened 2022-12-08
username logout part 1

armaniferrante:
closes https://github.com/coral-xyz/backpack/issues/1694

TODO:

- [x] UI
- [x] Figure out per user JWT switching

### DIFF
--- a/packages/app-extension/src/components/Locked/Reset/ResetWarning.tsx
+++ b/packages/app-extension/src/components/Locked/Reset/ResetWarning.tsx
@@ -22,6 +22,7 @@ export function Logout() {
   const background = useBackgroundClient();
   const nav = useNavStack();
   const user = useUser();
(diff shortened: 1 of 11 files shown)


tomlinton:
I think you can skip the JWT user switching. `WithAuth` handles it if the user changes, and the signing prompt will go away when the JWT cache is implemented.

armaniferrante:
Great news @tomlinton

armaniferrante:
Labeling as blocked on the jwt cache.

tomlinton:
JWT cache shouldn't be blocking, it'll work the same way as switching users with the extra signature prompt if its hardware. JWT cache will be merged/finished sometime this weekend though.

armaniferrante:
Going to do cookie deletion in the next pr.

PR #1777, opened 2022-12-08
async await persist fn

armaniferrante:


armaniferrante:
Followup loose end https://github.com/coral-xyz/backpack/pull/1779

Issue #1735, opened 2022-12-07
add local jwt cache


tomlinton:
The `WithAuth` wrapper component will reauth whenever the user switches but it'd be nice to cache JWTs locally for each user. Particularly true if a hardware wallet is involved because switching to a hardware account requires signing.

PR #1781, opened 2022-12-09
add jwt storage for users

tomlinton:
- Save the JWT in the store for each user.
- In the call to check if a user is authenticated in the `WithAuth` wrapper the backend will check the JWT in the header for authentication first, and if it's valid the cookie will be set (overwriting a possibly incorrect cookie if the cookie that exists has a JWT that is different from the JWT in the header).

Closes https://github.com/coral-xyz/backpack/issues/1735

### DIFF
--- a/packages/app-extension/src/hooks/useAuthentication.tsx
+++ b/packages/app-extension/src/hooks/useAuthentication.tsx
@@ -3,6 +3,7 @@ import {
   BACKEND_API_URL,
   UI_RPC_METHOD_KEYRING_STORE_LOCK,
   UI_RPC_METHOD_KEYRING_STORE_READ_ALL_PUBKEYS,
+  UI_RPC_METHOD_USER_JWT_UPDATE,
 } from "@coral-xyz/common";
 import { useBackgroundClient, useUser } from "@coral-xyz/recoil";
 import { ethers } from "ethers";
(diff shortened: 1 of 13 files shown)


packages/app-extension/src/hooks/useAuthentication.tsx
```diff
@@ -64,15 +67,14 @@ export const useAuthentication = () => {
...
-        {
-          method: "GET",
-          headers: {
-            "Content-Type": "application/json",
-          },
-        }
-      );
+      const headers = {
+        "Content-Type": "application/json",
+        ...(jwt ? { "Backpack-JWT": jwt } : {}),
```
armaniferrante, line 72: Unrelated to this PR, but should we use a more standard `Authorization: Bearer` header?

packages/background/src/backend/core.ts
```diff
@@ -841,10 +851,15 @@ export class Backend {
       const user = await store.getActiveUser();
       return user;
     } catch (err) {
-      return { username: "", uuid: "" };
+      return { username: "", uuid: "", jwt: "" };
     }
   }
 
+  async userJwtUpdate(uuid: string, jwt: string) {
+    await setUser(uuid, { jwt });
```
armaniferrante, line 859: We probably need to push out a user updated notification to the UI here.

packages/app-extension/src/hooks/useAuthentication.tsx
tomlinton: Yea agreed, I'll update it to something standard soon

PR #1787, opened 2022-12-10
use a single source of truth for active user uuid

tomlinton:
Any issue with passing the uuid in instead of grabbing it from the stored keyring JSON @armaniferrante? That way there is only a single source of truth for the active uuid.

I ran into a problem with the store and the json active user uuids getting out of sync (admittedly it was some dev foolery with the service worker dying so probably not a real bug but it seems like we could avoid this class of bugs altogether).

### DIFF
--- a/packages/background/src/backend/core.ts
+++ b/packages/background/src/backend/core.ts
@@ -636,7 +636,7 @@ export class Backend {
       throw new Error("invariant violation: username not found");
     }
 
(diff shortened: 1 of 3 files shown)


armaniferrante:
I can't think of a reason not to do this. For other readers, the single source of truth is the [here](https://github.com/coral-xyz/backpack/blob/master/packages/background/src/backend/store/usernames.ts#L5).

PR #1788, opened 2022-12-10
send active user updated notification on jwt update

armaniferrante:
Followup for https://github.com/coral-xyz/backpack/pull/1787

### DIFF
--- a/packages/background/src/backend/core.ts
+++ b/packages/background/src/backend/core.ts
@@ -857,6 +857,25 @@ export class Backend {
 
   async userJwtUpdate(uuid: string, jwt: string) {
     await setUser(uuid, { jwt });
+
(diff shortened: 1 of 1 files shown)


packages/background/src/backend/core.ts
```diff
@@ -857,6 +857,25 @@ export class Backend {
 
   async userJwtUpdate(uuid: string, jwt: string) {
     await setUser(uuid, { jwt });
+
+    const walletData = await this.keyringStoreReadAllPubkeyData();
+    const preferences = await this.preferencesRead(uuid);
+    const xnftPreferences = await this.getXnftPreferences();
+
+    this.events.emit(BACKEND_EVENT, {
```
armaniferrante, line 865: I don't think this notification is technically necessary. But it will be if we ever need the most up to date JWT in the front end `user` atom.

PR #1795, opened 2022-12-11
dumb fix for ethereum

peterpme:


packages/background/src/backend/ethereum-connection.ts
```diff
@@ -25,6 +25,14 @@ const logger = getLogger("ethereum-connection-backend");
 export const ETHEREUM_TOKENS_REFRESH_INTERVAL = 10 * 1000;
 export const ETHEREUM_FEE_DATA_REFRESH_INTERVAL = 20 * 1000;
 
+// defaultProvider breaks mobile
+function handleResponse(response) {
```
armaniferrante, line 29: Is this really on all of these methods?

armaniferrante: Also it seems better to put this into `frontend/ethereum-connection`?

PR #1799, opened 2022-12-11
fix show secret phrase on keyring type change

tomlinton:


armaniferrante:
Your solution is much more elegant than mine https://github.com/coral-xyz/backpack/pull/1797/files

PR #1829, opened 2022-12-13
Profile Pictures

greenish:


packages/background/src/backend/core.ts
```diff
@@ -610,6 +610,35 @@ export class Backend {
     return SUCCESS_RESPONSE;
   }
 
+  async userAvatarUpdate(uuid: string): Promise<string> {
```
armaniferrante, line 613: Is this needed still?

greenish: nope -> removed

packages/app-extension/src/components/Unlocked/Nfts/Detail.tsx
```diff
@@ -27,6 +27,7 @@ import MoreHorizIcon from "@mui/icons-material/MoreHoriz";
 import { IconButton, Popover, Typography } from "@mui/material";
 import { PublicKey } from "@solana/web3.js";
 import { BigNumber } from "ethers";
+import { relative } from "path";
```
wentokay, line 30: maybe an accidental auto-import?

packages/common-public/src/utils.ts
```diff
@@ -58,12 +58,19 @@ export function externalResourceUri(uri: string): string {
...
 
-export function proxyImageUrl(url: string): string {
+export function proxyImageUrl(
+  url: string,
+  maxWidth = 400,
+  maxHeight = 400
+): string {
   if (url && (url.startsWith("http://") || url.startsWith("https://"))) {
-    return `${IMAGE_PROXY_URL}/insecure/rs:fit:400:400:0:0/plain/${url}`;
+    return `${IMAGE_PROXY_URL}/cdn-cgi/image/fit=contain,width=${maxWidth},height=${maxHeight},quality=85/${url}`;
```
wentokay, line 73: I might missing some context here but if you make this change you might also want to change `IMAGE_PROXY_URL` to https://images.xnfts.dev so that we don't need to do the redirect from cloudfront>cloudflare? cc @hkirat

PR #1837, opened 2022-12-13
restart autolock countdown after receiving UI_RPC_* messages

wentokay:
fixes #1393 based on @tomlinton (& @armaniferrante's) [suggestion](https://github.com/coral-xyz/backpack/pull/1478#discussion_r1021011451)

# todo

- [ ] disable autolock if an xnft is open
- [ ] grok `lastUsedTs` to check if it can be removed or ignored
- [ ] verify that only UI_RPC methods are included and polling updates are excluded
- [ ] verify this works ok with multiple user profiles, will each username have its own timeout setting?
- [ ] more testing

PR #1858, opened 2022-12-15
autolock timer fixes and updates

wentokay:
replaces #1837 

- [ ] grok lastUsedTs to check if it can be removed or ignored
- [ ] verify that only UI_RPC methods are included and polling updates are excluded
- [ ] verify this works ok with multiple user profiles, will each username have its own timeout setting?
- [ ] check that this doesn't break anything with `app-mobile`
- [ ] use a more robust approach to check if xnft is open or closed? Or add comment where route is defined

## scenarios to test

status | scenario | expected behavior
----|-----|----
✅ | user opens xnft | should not lock
✅ | user opens xnft and closes xnft | should lock after X minutes of inactivity
✅ | user opens backpack | should lock after X minutes of inactivity
✅ | user opens backpack and closes it | should lock after X minutes of inactivity
✅ | user changes autolock time from X to Y minutes | should lock after Y minutes of inactivity
🔜 done, needs more testing | user opens xnft and closes popup | should lock after X minutes of inactivity
🔜 | user opens xnft and closes popup | should lock immediately
🔜 | user unlocks backpack with xnft auto-opened | should not lock

* inactivity is defined as the interval of time since the last UI_RPC call


package.json
```diff
@@ -18,7 +18,7 @@
     "start": "env-cmd --silent turbo run start --concurrency=100% --filter=./packages/*",
     "start:fresh": "yarn install && yarn clean && yarn install && yarn start",
     "test": "env-cmd --silent turbo run test -- --passWithNoTests --watchAll=false",
-    "build": "env-cmd --silent turbo run build",
+    "build": "env-cmd --silent turbo run build --filter=@coral-xyz/recoil && env-cmd --silent turbo run build",
```
wentokay, line 21: necessary because of circular dependencies in the project, will try to fix but it might be out of scope for this PR

package.json
```diff
@@ -34,7 +34,7 @@
     "eslint-plugin-react": "^7.30.1",
     "husky": "^8.0.1",
     "lint-staged": "^12.4.1",
-    "prettier": "^2.7.1",
+    "prettier": "^2.8.1",
```
wentokay, line 37: needed for [`satisfies`](https://prettier.io/blog/2022/11/23/2.8.0.html#support-typescript-49-satisfies-operatorhttpsdevblogsmicrosoftcomtypescriptannouncing-typescript-4-9satisfies-13764httpsgithubcomprettierprettierpull13764-13783httpsgithubcomprettierprettierpull13783-13872httpsgithubcomprettierprettierpull13872-by-sosukesuzukihttpsgithubcomsosukesuzuki)

packages/background/src/backend/keyring/index.ts
```diff
@@ -65,6 +71,93 @@ export class KeyringStore {
...
+          // Force-enable the auto-lock countdown if the popup is closed
+          autoLockIsEnabled = true;
+          if (shouldLockImmediatelyWhenClosed) {
+            lockImmediatelyWhenClosedCountdown = setTimeout(() => {
+              stopAutoLockCountdownTimer();
+              lock();
+            }, SECONDS_UNTIL_LOCK_WHEN_CLOSED * 1000);
+          }
+        });
+      });
```
wentokay, line 120: @peterpme if you get chance could you do me a huge favor and check if this branch (`wen/autolock`) breaks any mobile stuff for you please? I'm particularly concerned about this extra `chrome.runtime` stuff in the background script. I think I might need to add something to make it react-native-friendly

armaniferrante:
To answer your question, yes every user has a separate set of preferences that are currently stored client side by UUID and so will have their own timeout settings. https://github.com/coral-xyz/backpack/blob/master/packages/background/src/backend/store/preferences.ts#L13

packages/app-extension/src/index.tsx
```diff
@@ -7,6 +7,9 @@ import "./index.css";
 const App = lazy(() => import("./app/App"));
 const LedgerIframe = lazy(() => import("./components/LedgerIframe"));
 
+// Connect to the background script so it can detect if the popup is closed
+chrome.runtime.connect();
```
peterpme, line 11: i think as long as this doesn't show up in any common code, it should be ok! i wrote this script previously: https://github.com/coral-xyz/backpack/blob/master/packages/common/src/channel/app-ui.ts#L35-L41

wentokay:
> To answer your question, yes every user has a separate set of preferences that are currently stored client side by UUID and so will have their own timeout settings. https://github.com/coral-xyz/backpack/blob/master/packages/background/src/backend/store/preferences.ts#L13

ok, after a quick initial test if I have user A and user B with 15 min default autolock preference, then I change user A's preference to 1 minute and switch account to user B it's locking in 1 minute, when it should be 15 minutes

wentokay:
the final issue is that if I set the extension to `never` lock, then quit and reopen the browser, backpack locked because the background script was forced to close. I guess the preference would need to be persisted in some way outside the encrypted store if this was going to be a lasting option

wentokay:
> > To answer your question, yes every user has a separate set of preferences that are currently stored client side by UUID and so will have their own timeout settings. https://github.com/coral-xyz/backpack/blob/master/packages/background/src/backend/store/preferences.ts#L13
> 
> ok, after a quick initial test if I have user A and user B with 15 min default autolock preference, then I change user A's preference to 1 minute and switch account to user B it's locking in 1 minute, when it should be 15 minutes

I'm going to try and address this by resetting the auto-lock timer when `UI_RPC_METHOD_ACTIVE_USER_UPDATE` is received

packages/background/src/backend/store/preferences.ts
```diff
@@ -13,7 +18,8 @@ const STORE_KEY_WALLET_DATA = "wallet-data";
 export type WalletData = {
   // Deprecated. Don't use this.
   username?: string;
-  autoLockSecs: number;
+  autoLockSecs?: ExtractRecoilType<typeof autoLockSecs>;
+  autoLockOption?: ExtractRecoilType<typeof autoLockOption>;
```
armaniferrante, line 22: Instead of using a separate variable, would it make more sense to combine this into `autoLockSecs` ?

wentokay: Yep, no problem

My reasoning was -

> this adds autoLockOption: "onClose" | "never" | undefined to the store, rather than change autoLockSecs: number, it does this to keep the migration smooth for existing users and not break compatibility with new users
> so existing users will still have autoLockSecs=(15*60) by default, and autoLockOption will be undefined

I'll combine them and add a migration path instead

package.json
wentokay: I'm going to remove some types (and their imports) I added to avoid this issue for now, it's out of scope

PR #1873, opened 2022-12-16
Iframe-ify messaging

hkirat:
The PR
1. Moves all messaging into `message-sdk` and exposes it as a react app
2. Moves common stuff b/w `message-sdk` and `app-extension` into `react-common`
3. Replaces the `messages` tab with an `iframe` to `http://localhost:3000` (this will be changed to `messages.xnfts.gg` after the PR id deployed)
4. Cuts down some depedencies from `chat-sdk
5. Adds feature gates to flip b/w iframe and non iframe version


package.json
```diff
@@ -46,7 +46,11 @@
       "eslint --fix"
     ]
   },
+  "overrides": {
```
hkirat, line 49: Addded this to make a `react app` work inside the mono repo. Suggested here - https://github.com/facebook/create-react-app/issues/11810

PR #1882, opened 2022-12-16
Add swr cache for offchain NFT metadata fetching

greenish:


packages/recoil/src/hooks/solana/useLoadSplTokens.tsx
```diff
@@ -15,12 +15,16 @@ export const useUpdateAllSplTokenAccounts = () =>
...
+        const atom = atoms.customSplTokenAccounts({ connectionUrl, publicKey });
+
+        const currentValue = get(atom);
+        const newValue = {
           ...customSplTokenAccounts,
           splTokenMints: new Map(customSplTokenAccounts.mintsMap),
-        });
+        };
+
+        if (JSON.stringify(currentValue) !== JSON.stringify(newValue)) {
```
armaniferrante, line 26: Will this always be true because of the map not having a defined ordering?

greenish: Oh.. I didnt mean to push that here, this was just a test and will come later.

greenish: fixed

PR #1898, opened 2022-12-17
wallet selector on origin approval

armaniferrante:


armaniferrante:
Note that this still needs design work.

Issue #1879, opened 2022-12-16
solana priority fees


armaniferrante:
dev mode only

armaniferrante:
Details on how to do this here https://solanacookbook.com/references/basic-transactions.html#how-to-change-compute-budget-fee-priority-for-a-transaction

PR #1900, opened 2022-12-17
adds solana priority fee

hkirat:
Adds an option to provide `priority fee` constructs if `developer mode` is enabled.

Closes https://github.com/coral-xyz/backpack/issues/1879

New designs (Related https://github.com/coral-xyz/design-issues/issues/40)
Screenshot 2022-12-19 at 4 30 00 PM
Screenshot 2022-12-19 at 4 57 23 PM


packages/app-extension/src/components/Unlocked/ApproveTransactionRequest.tsx
```diff
@@ -466,6 +471,11 @@ function SendTransactionRequest({
               transactionData={transactionData}
             />
           </div>
+          {blockchain === "solana" && (
```
tomlinton, line 474: Can this be combined with the already existing structure for user configurable parameters for Ethereum here? https://github.com/coral-xyz/backpack/blob/master/packages/app-extension/src/components/common/TransactionData.tsx#L109

aarongillett:
Some UI updates to this in case it’s useful. Key points:

- Denser metadata. There’s now a lot of info here so best if it all fits without scrolling.
- Inline inputs (assuming compute units and priority fee are both editable inputs)
- Number formatting for inputs would help legibility if big numbers are expected.
- Might help to be more specific with _Max_. Perhaps _Max fee_?
- Is it ok that _Max_ is in $ or is the SOL balance also useful? If so, we could format like `0.00003 SOL ($0.00006)`
- Should the error message be after the metadata as a summary, or does your version make more sense?

image

aarongillett:
All xNFT tx approvals should be styled the same as general extension txs for consistency and so we don’t need to maintain two versions.

The xNFT action buttons should probably also be the same (Deny/Approve) for consistency.

aarongillett:
There should be inline inputs for ETH gas settings, so you might be able to recycle that component.

hkirat:
Transaction done from `examples/clients/simple` => https://explorer.solana.com/tx/KBmRVg4qicLtbX24QxVmY49F5Qfd74CDH2qmooUitUR9WZRU7tUJXuP8gBajbZygeCz2YxJ2ZEX9Mt2qVNxtWwv
Transaction done from `examples/xnft/provider-test` => https://explorer.solana.com/tx/3E3rKG3umFzaZV5VX93fYqi7B5yjtQUXDL52sigBKWteZgqceEfwMjwRNtd5ZL2D3obo87NJ6gTCMgE6QmeeTA8u

hkirat:
Fee deducted 0.010005 
Screenshot 2022-12-19 at 4 57 23 PM
Screenshot 2022-12-19 at 4 56 00 PM

hkirat:
This isn't working with NFT marketplaces, I think because they are checking transaction parameters after the tx is signed and we are adding instructions to the txns when `priority fee` is enabled by the user.
Here's the error that pops up on magiceden - 
Screenshot 2022-12-19 at 6 33 23 PM

We might have to get in touch with them for them to allow these

Issue #1917, opened 2022-12-19
Update footer on avatar edit screen to use correct theme


aarongillett:
In dark mode, the footer is using light mode.
![image](https://user-images.githubusercontent.com/4157473/208473954-245ac596-6d28-4acf-b03c-e40f798466fb.png)

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1927.

Issue #1906, opened 2022-12-18
useAvatarUrl hook is not user specific


tomlinton:
Once you update an avatar the newAvatar atom is set in recoil, and that sets it for all accounts because if it exists it is used instead of querying based on username.

https://github.com/coral-xyz/backpack/blob/d12f6c73a91cd0a314d7bec549be3d98b1148e63/packages/recoil/src/hooks/avatar.tsx#L11

<img width="376" alt="Screenshot 2022-12-18 at 4 37 14 PM" src="https://user-images.githubusercontent.com/489202/208280164-a82ad184-6129-49fc-9710-18e596c3f915.png">

armaniferrante:
@greenish any ideas on this one?

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1927.

Issue #1888, opened 2022-12-17
pfp select screen periodically jumps to top


tomlinton:
Open the PFP selection screen. Scroll down (you need enough NFTs in the wallet to make it scrollable). It'll jump back to the top after some time. This likely happens when the token polling update hits and the recoil state is updated.

armaniferrante:
I suspect this will be addressed by the work @greenish is doing to improve NFT load perf.

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1933.

Issue #1870, opened 2022-12-16
Collectibles page breaks


armaniferrante:
Steps to reproduce:

* open backpack
* change to nft tab
* change to balances tab
* change back to nft tab

<img width="376" alt="Screen Shot 2022-12-16 at 12 27 03 PM" src="https://user-images.githubusercontent.com/6990215/208154586-2f7fbc07-550f-41a4-a79c-52915e59ed31.png">

armaniferrante:
Probably caused by https://github.com/coral-xyz/backpack/pull/1860

armaniferrante:
closed by https://github.com/coral-xyz/backpack/pull/1877

tomlinton:
Still present on build #1428

greenish:
I used to be able to reproduce this by just changing tabs before #1877 but not able to anymore.. Are there any other steps to cause this?

tomlinton:
It seems to happen when I do anything. Maybe try with multiple blockchains holding NFTs.

https://user-images.githubusercontent.com/489202/208398226-e92c17c9-86fe-44d1-b038-75a0f31961e6.mov

armaniferrante:
Closed by https://github.com/coral-xyz/backpack/pull/1933.

PR #1927, opened 2022-12-19
Avatar cache tweaks & hook update

greenish:
this should fix 
- #1906
- #1888
- #1917

And a potential fix for #1870 -> I cant reproduce this issue, my theory is that specific NFTs cause this.


packages/themes/src/index.tsx
```diff
@@ -318,7 +318,7 @@ export const lightTheme: Partial<Theme> & { custom: { colors: CustomColors } } =
         secondary: LIGHT_FONT_COLOR_1,
         primaryButton: LIGHT_BRAND_COLOR,
         primaryButtonTextColor: LIGHT_BACKGROUND_COLOR_1,
-        secondaryButton: LIGHT_BACKGROUND_COLOR_1,
+        secondaryButton: LIGHT_BACKGROUND_COLOR_2,
```
armaniferrante, line 321: Seems like this might regress colors elsewhere.

armaniferrante: Confirmed. For example, the cancel button on the send NFT screen.

armaniferrante, line 321:
```suggestion
        secondaryButton: LIGHT_BACKGROUND_COLOR_1,
```

armaniferrante:
Colors look good even after merging.

armaniferrante:
Note that this did not solve https://github.com/coral-xyz/backpack/issues/1888.

greenish:
> Note that this did not solve #1888.

Ah this should have been #1889 

GH autocomplete very easily messes up

PR #1932, opened 2022-12-20
track referrals

wentokay:
a user visits a referral link that includes the referrer's `:username`

the auth worker looks up the referrer `uuid` using the `:username`, then sets a (referrer=`uuid`) cookie on `*.xnfts.dev` and redirects the user to the chrome store download page

the user downloads the extension, signs up and if the referrer `uuid` cookie exists, it is added to the new user record (user.referrer_id)

## todo

- [x] add a route to track the referrer, set the cookie and redirect the user to chrome store
- [x] read referrer cookie inside `POST /users` signup method
- [x] clear referrer cookie after signup
- [x] manual tests

## future ideas

- use a short url for referrals?
- scope the referrer cookie to the auth path?
- if the user already has backpack installed, show them a page where they can create/share their own referral links instead of redirecting to backpack?
- dashboards for tracking referrals
- add ability to create custom referral links? for different campaigns?
- decide what to do if the user has a previously shared referral link and changes their username
- decide if it's necessary to track the referral click internally? if so should the new user's uuid be created and stored in a cookie at that time too, so that it can be deterministically tracked before the new record is created


backend/native/backpack-api/src/routes/v1/referrals.ts
```diff
@@ -0,0 +1,19 @@
...
+import { CHROME_STORE_URL, REFERRER_COOKIE_NAME } from "../../config";
+import { getUserByUsername } from "../../db/users";
+
+const router = express.Router();
+
+/**
+ * Set a cookie on *.xnfts.dev that will be used to track referrals, then
+ * redirect to the backpack chrome extension page
+ */
+router.get("/:username", async (req, res) => {
```
armaniferrante, line 13: What happens if I visit such a link *after* signing up? Visiting the referral link shouldn't do anything in such a case.

armaniferrante: On second thought, it should be fine to set the cookie here, assuming it's only ever used during onboarding. If the cookie exists during onboarding, then use it to set the referral in the backend and use that in the future, ignoring the cookie till the end of time.

wentokay: could you elaborate on not doing anything, what should be shown there in this case?

wentokay: I considered this in notes, but thought it might be out of scope?

> if the user already has backpack installed, show them a page where they can create/share their own referral links instead of redirecting to backpack?

armaniferrante: We can just redirect. Agreed it's out of scope and low priority to do anything else.

wentokay: OK. The `referrer=${uuid}` cookie gets cleared by the API after it's used during a successful signup as things are currently. I'm still not sure if that was a sensible idea or not 😅

armaniferrante: Sounds sensible to me!

wentokay: Ah I see what you mean now, your concern was that an existing user could potentially have a different referral URL associated with them if they visited the link after signing up.

Yeah so it'd be ignored right now, currently the cookie would expire on its own after 1 year but that could be shortened.

Another thing I considered was storing the cookie as something like

`referrer={ newUserId: uuid(), referrerId: referrer.id }` then when the user signs up it uses grabs the provided `uuid()` as the new user ID, but I don't know if that provides much extra value right now

wentokay:
when this gets merged hasura migrations would ideally want to be applied before the API is autodeployed by github actions

PR #1933, opened 2022-12-20
New nftCollections Atoms (& remove useLoader)

greenish:
- New Collections atom that uses effects to decouple polling updates from UI
- Update Avatar dashed border radius
- Removes need of useLoader hook and fixes 
  - #1888 
  - #1870


armaniferrante:
The nft tables are no longer respecting whether a blockchain has been disabled.

greenish:
> The nft tables are no longer respecting whether a blockchain has been disabled. I.e., a loading table displays for ethereum even if disabled.

Should be fixed 👍

PR #1475, opened 2022-11-13
merge common-public back into common

wentokay:
removes common-public by merging all of its code into common

PR #1955, opened 2022-12-22
merge common-public into common

wentokay:
removes `@coral-xyz/common-public` by merging all of the code back into `@coral-xyz/common`

replaces #1475


wentokay:
not sure about some of these yarn.lock changes... looking into it

wentokay:
ah, still blocked because the published npm `react-xnft` has `@coral-xyz/common-public` as a dependency

wentokay:
closing again as it's going to be stale, will return soon

PR #1977, opened 2022-12-31
Updated dockerfiles for deploying backend services

hkirat:


packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Solana/index.tsx
```diff
@@ -310,7 +310,7 @@ export const isCardinalWrappedToken = async (
   try {
     programs.transferAuthority.accounts.getTransferAuthority(
       connection,
-      tokenManagerData?.parsed.transferAuthority
+      tokenManagerData?.parsed.transferAuthority || new PublicKey("")
```
hkirat, line 313: There is a type error here `PublicKey | null is not assignable to PublicKey` causing the backend builds to fail, hence added a default here.

PR #1665, opened 2022-12-02
Add nohoist to examples/**

greenish:
This PR adds `nohoist` to all `example/` packages:

The goal is to continue to make use of yarns package linking (installing local packages as dependencies) while managing dependencies independently.

- Keeping examples standalone / closer to real life.
- Enabling examples to have iE outdated dependencies without affecting the rest of the workspaces.

Also adding `nohoist` to `app-mobile` since `react-native` dependencies are sensitive and the dependency trees between mobile and the rest of the app have larger differences so hoisting becomes less valuable.

PR #1984, opened 2023-01-02
fix nohoist

peterpme:
TLDR nohoist doesn't work on workspace folders, just on dependencies

read the full thing:

https://github.com/coral-xyz/backpack/pull/1665#issuecomment-1369050100


peterpme:
with nohoist turned on, this seems to be breaking the mobile app. `@coral-xyz/recoil` doesn't work anymore :(

PR #2016, opened 2023-01-05
update: remove hover for text element

### LINKED ISSUE #2013: Remove hover state from wallet connect request
The two components “View wallet balances…” and “Request approval…” both have a hover state, but shouldn’t (they’re just text, not interactive).

This hover state may also be on other wallet connect/tx approval screens.

image

jim4067:
resolves #2013

### DIFF
--- a/packages/app-extension/src/components/Unlocked/Approvals/ApproveOrigin.tsx
+++ b/packages/app-extension/src/components/Unlocked/Approvals/ApproveOrigin.tsx
@@ -48,6 +48,9 @@ const useStyles = styles((theme) => ({
     padding: "8px",
     marginBottom: "1px",
     border: `${theme.custom.colors.borderFull}`,
+    "&:hover": {
+      background: "#fff !important",
+    },
   },
   listItemIconRoot: {
     minWidth: "inherit",
--- a/packages/app-extension/src/components/Unlocked/Approvals/ApproveOrigin.tsx
+++ b/packages/app-extension/src/components/Unlocked/Approvals/ApproveOrigin.tsx
@@ -49,7 +49,8 @@ const useStyles = styles((theme) => ({
     marginBottom: "1px",
     border: `${theme.custom.colors.borderFull}`,
     "&:hover": {
-      background: "#fff !important",
+      background: `${theme.custom.colors.hoverBackground} !important`,
+      opacity: "1",
     },
   },
   listItemIconRoot: {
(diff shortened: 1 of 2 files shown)


packages/app-extension/src/components/Unlocked/Approvals/ApproveOrigin.tsx
```diff
@@ -48,6 +48,9 @@ const useStyles = styles((theme) => ({
     padding: "8px",
     marginBottom: "1px",
     border: `${theme.custom.colors.borderFull}`,
+    "&:hover": {
+      background: "#fff !important",
```
armaniferrante, line 52: I think this will not do what we want in dark mode?

aarongillett: I think you’re right. We want to remove the hover state, and this looks like it hard-codes one instead.

jim4067: thanks. hadn't thought of that.

@aarongillett that's not possible for material design elements.

packages/themes/src/index.tsx
```diff
@@ -114,6 +114,7 @@ type CustomColors = {
   coralGradient: string;
   miniDrawerBackdrop: string;
   unreadBackground: string;
+  hoverBackground: string;
```
armaniferrante, line 117: Can we do this without introducing another field here?

jim4067: I don't think so, 

The other option I thought of would be to explicitly introduce the `palette: {mode: 'light' | 'dark' }` field in the light and dark theme definitions and using this conditionally determine the colour of the `li` element when hovered.

```ts

export const darkTheme: Partial<Theme> & { custom: { colors: CustomColors } } =
  {
    ...baseTheme,
    components: darkComponentOverrides,
    palette: {
      mode: 'dark'
    },

...

export const lightTheme: Partial<Theme> & { custom: { colors: CustomColors } } =
  {
    ...baseTheme,
    components: lightComponentOverrides,
    palette : {
      mode: "light"
    },

//do conditional rendering of which color to display on hover
marginBottom: "1px",
    border: `${theme.custom.colors.borderFull}`,
    "&:hover": {
      background: `${theme.palette?.mode === "light" ? theme.custom.colors.brandColor: theme.custom.colors.banner}`,
      opacity: "1",
    },

PR #2029, opened 2023-01-09
Adds secure transfer as a message type

hkirat:
Uses https://github.com/coral-xyz/secure-transfer to allow users to escrow money and wait for the other party to accept before it directly gets debited from them.
Screenshot 2023-01-10 at 11 43 54 PM
Screenshot 2023-01-11 at 10 58 08 PM
Screenshot 2023-01-11 at 10 57 49 PM
Screenshot 2023-01-11 at 10 51 34 PM


aarongillett:
@hkirat The secure transfer modal; is that what is displayed when someone selects the $ icon in the chat input bar?
This is a good start, but I think we need a few more pieces, like education around what secure send means. I can work on this tomorrow and get you some designs hopefully by the end of the day to review.

jokumusic:
I was about to start implementing this, good thing I checked first. I think the 3 way handshake will work for what I need, but I was originally thinking of an escrow account that contains a withdraw hash where the withdrawer sends a password, the program hashes the password and compares it to the withdraw hash in the escrow account.

PR #2220, opened 2023-01-24
Auto enable ONE nft notifications

greenish:
Sets default `pushNotifications` for ONE nft to true


hkirat:
We might have to add a whitelist here as well - https://github.com/coral-xyz/backpack/blob/master/backend/native/xnft-api-server/src/controllers/notifications.ts#L43 
Else the backend wont send the push notification

Issue #2214, opened 2023-01-24
Dark theme font color in messages


callensm:
"Send Request" in dark mode is black text.

<img width="365" alt="image" src="https://user-images.githubusercontent.com/13121516/214355022-3af5e63b-882c-466f-a4ea-a63b39cba0eb.png">

jim4067:
@callensm, should they be white?

PR #2221, opened 2023-01-24
send request color fix in dark mode

harishlalwani:


harishlalwani:
https://github.com/coral-xyz/backpack/issues/2214 

@callensm - check if this works

Issue #753, opened 2022-09-05
transaction balance change estimation doesn't handle native SOL


tomlinton:
e.g. swap from SOL to DUST

<img width="374" alt="Screen Shot 2022-09-05 at 12 15 26 PM" src="https://user-images.githubusercontent.com/489202/188487915-7369d74f-4b35-475d-b0ba-4f50df797d72.png">

PR #2230, opened 2023-01-25
balance change simulation improvements (sol and versioned transaction support)

tomlinton:
- Adds native SOL support to simulation
- Adds versioned transaction support by looking up the correct keys and passing them to simulate
- Fixes a bug where only the wallet public key was passed to simulate so balance changes to ATAs would not be found

Closes https://github.com/coral-xyz/backpack/issues/753


POPPIN-FUMI:
@tomlinton Thank you for fixing this issue!
I am Kawasaki from EpicsDAO,
We empower open-source software development.
This quest is related to Epics Alpha's quest.
So you get 6000$EPCT as a reward if you submit this PR link to this platform
https://alpha.epics.dev/en/quest/UXVlc3Q6MTE5/

Thank you again!

PR #2246, opened 2023-01-25
add missing alter_table command

greenish:


wentokay:
I think this needs to run before the recent create view migrations, it might be possible to do it just by changing the timestamp in the directory name to a number that is lower than them, but not 100% sure if that is enough

greenish:
Was able to deploy this locally 🤞

PR #2257, opened 2023-01-26
pNFT transfers

crypt0miester:
still draft as the code is untested.

following changes has been done: 
1. update `@metaplex-foundaiton/mpl-token-metadata` to latest in app-extension and common
2. add `findTokenRecordPda` in common
3. add transferProgrammableNonFungible

need to mint a pNFT to test. please hold on until I lift the draft PR.

thanks.


crypt0miester:
two possible solutions I can think of: 
1. update mpl-token-metadata dependency across packages.
2. extract needed classes and functions from mpl-token-metadata since we only need the transfer from the whole update.

please inform me which would you prefer me doing? perhaps something else that I have not thought of?

PR #2273, opened 2023-01-27
deletes unused service-worker-loader directories that break the build bc its 9gb?

peterpme:
## problem

TIL that every commit hash is saved into a directory under the `gh-pages` branch

none of the old directories are ever deleted so we're left with a branch with over 2k folders (one for every commit against master)

why is this a problem? bc it seems like our builds have started failing bc it's trying to upload 9gb of data (seemingly the whole directory of folders)

unless there's something else i'm missing... i've deleted all of these folders to clean the repo up

## next steps

we should permanently delete the old commits from here since there's no value for them right now. there's a special git command we can run

## moving forward

- every couple of weeks remember to delete old folders OR
- write a script that deletes everything except for the latest 7 days worth of commits AND
- point production to the latest commit hash bc this'll be a problem once we go live


wentokay:
Sweet, thanks @peterpme!

The build action was added very early on when there wasn't much activity happning in the repo, I feel bad for the azure workers trying to compile this stuff now 😆

It'd be good to add a scheduled action to delete old dirs like you say. Lazy question, why is it creating an artifact of the multi-gigabyte branch, shouldn't it just be copying a .js and .html from the main build?

A couple of things to add on the potential ideas list -

- only build a new service worker if mobile app code has changed in the PR?
- deploy the code to S3 or R2 instead? Github pages was always a convenience thing

PR #2350, opened 2023-01-31
Notifications onboarding and settings

### LINKED ISSUE #2326: Notifications polish
- Move notifications toggle
- Add toggle to preferences
- Update notifications empty state

[Figma](https://www.figma.com/file/HNF2HJZYq6dUzCWxSmH2Oe/Backpack-_-Questlog?node-id=17267%3A130540&t=tPI9KcXlC1aNHMIf-1)

callensm:
closes #2326 
Screenshot 2023-01-30 at 7 03 07 PM

### DIFF
--- a/packages/app-extension/src/components/Onboarding/pages/NotificationsPermission.tsx
+++ b/packages/app-extension/src/components/Onboarding/pages/NotificationsPermission.tsx
@@ -1,0 +1,98 @@
+import { useState } from "react";
+import { PrimaryButton, SecondaryButton } from "@coral-xyz/react-common";
+import { useCustomTheme } from "@coral-xyz/themes";
+import { Box } from "@mui/material";
+
+import {
+  registerNotificationServiceWorker,
+  saveSubscription,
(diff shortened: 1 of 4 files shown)


callensm:
@trucy @armaniferrante i'm not sure notifications can work as a switch type of preference. enabling from preferences would be fine if not already done, but an application cannot mutate browser API permissions and there is no programmatic method (that is still supported by browser APIs) to revoke permissions.

armaniferrante:
@callensm forwarding from @hkirat. We can show the switch in preferences, and if it's disabled, clicking it on should take the user to /permissions?notifications=true (we can’t turn them on from a popped out extension, need it open in a browser tab) which should do it.

hkirat:
@callensm we can delete the users token from our DB when they turn off notifications. I'll expose an endpoint that you can hit when the user decides to disable notifications. 
If we don't have their token, we wont be able to send them notifications (even if they've given native browser permissions)

PR #2358, opened 2023-01-31
fix notification screen crash

armaniferrante:


packages/app-extension/src/components/common/Layout/Router.tsx
```diff
@@ -104,6 +108,13 @@ export function Redirect() {
   return <Navigate to={url} replace />;
 }
 
+// We use a separate redirect for the xs size because some routes, e.g., /notifications
+// and /recent-activity don't exist on the xs size--for xs, they are ephemeral drawers,
+// for larger screens they are normal routes.
+export function RedirectXs() {
```
armaniferrante, line 114: Note: the alternative to this would be to make drawers in backpack persistent routes.

Issue #2380, opened 2023-02-01
clicking app signing toggle sends two rpc requests


tomlinton:
It is impossible to disable because it sends one false then one true.

<img width="799" alt="Screenshot 2023-02-01 at 11 16 01 PM" src="https://user-images.githubusercontent.com/489202/216015369-5ad4dbfe-760f-408d-9e12-7462976697f8.png">

PR #2383, opened 2023-02-01
fix is cold bug

tomlinton:
Fix for https://github.com/coral-xyz/backpack/issues/2380 but untested. Will test in morning if unmerged!


packages/app-extension/src/components/Unlocked/Settings/YourAccount/EditWallets/WalletDetail.tsx
```diff
@@ -135,10 +135,10 @@ export const WalletDetail: React.FC<{
       detail: (
         <ModeSwitch
           enabled={!isCold}
-          onSwitch={async (enabled) => {
+          onSwitch={async () => {
```
armaniferrante, line 138: Note: the reason why I added back in the onClick handler above is that the switch needs to toggle regardless of where you click.

PR #2393, opened 2023-02-01
Store images in localstorage for faster lookups

hkirat:
PR stores images in the localstorage from where we lookup before hitting the server. 
This causes the image loads to be much faster


hkirat:
Downside of this approach is that gifs are no longer dynamic (they become images)

Issue #1782, opened 2022-12-09
useMetadata hook doesn't return anything in react-native (xnft-quickstart)


hkirat:
Seems to be working in the `non react native` examples

PR #2395, opened 2023-02-01
provide window.xnft.metadata to iframe & native xNFTs

wentokay:
please see comment below before merging

### DIFF
--- a/packages/provider-core/src/root-provider-xnft.ts
+++ b/packages/provider-core/src/root-provider-xnft.ts
@@ -121,6 +121,13 @@ export class ProviderRootXnftInjection extends PrivateEventEmitter {
       );
     }
 
+    if (this.#cachedNotifications[PLUGIN_NOTIFICATION_UPDATE_METADATA]) {
+      iframeEl.contentWindow?.postMessage(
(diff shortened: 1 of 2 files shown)


packages/provider-core/src/root-provider-xnft.ts
```diff
@@ -121,6 +121,13 @@ export class ProviderRootXnftInjection extends PrivateEventEmitter {
       );
     }
 
+    if (this.#cachedNotifications[PLUGIN_NOTIFICATION_UPDATE_METADATA]) {
+      iframeEl.contentWindow?.postMessage(
+        this.#cachedNotifications[PLUGIN_NOTIFICATION_UPDATE_METADATA],
+        "*"
+      );
+    }
```
wentokay, line 129: are there any negative ramifications from doing this? In theory could it pass the metadata object to any react-xnft  `<Iframe xnft />` element?

wentokay:
related #1782

packages/provider-core/src/root-provider-xnft.ts
greenish: I don't think thats a problem since it still means its running inside backpack somewhere in the parent so I dont see a problem passing it down further.

PR #2440, opened 2023-02-02
Remove xnft-wrapper proxy -> not necessary with html xnfts anymore

greenish:


armaniferrante:
Will this break existing xnfts like tiexo?

wentokay:
Would xnfts still be hosted on embed.xnfts.dev? I think this is necessary for including the JWT in requests to .xnfts.dev domains

greenish:
> Will this break existing xnfts like tiexo?

With the new protocol, all entrypoints are html entrypoints so can be loaded directly.

> Would xnfts still be hosted on embed.xnfts.dev? I think this is necessary for including the JWT in requests to .xnfts.dev domains

hm no, xnfts wouldnt be hosted on embed.xnfts.dev anymore.

With the new protocol we can have multi-file xnfts which resolve relative urls. 
These relative urls dont work through the current proxy. Same for "iframe" xnfts with external urls.

so we might need a better proxy that can resolve relative paths as well.

armaniferrante:
You're right. I forgot @callensm had migrated the existing xNFTs.

greenish:
@wentokay this PR adds the userId to the window.xnft.metadata object. Would that remove the need for the jwt in xnfts for you? https://github.com/coral-xyz/backpack/pull/2444

PR #2353, opened 2023-01-31
Typo fix

i-m-aditya:

PR #2471, opened 2023-02-04
Truncate long wallet names

### LINKED ISSUE #2462: Long wallet names should truncate
Let’s truncate long names with the format
`Ledger Nano X Bac…p`

image

Two instances (maybe more in other places)

Balances table view
image

Wallet switcher drawer
image

i-m-aditya:
Fixes #2462

### DIFF
--- a/packages/app-extension/src/utils/util.ts
+++ b/packages/app-extension/src/utils/util.ts
@@ -1,0 +1,6 @@
+export const truncateName = (name: string, size: number) => {
+  if (name.length > size) {
+    return name.slice(0, size - 4) + "..." + name.slice(-1);
+  }
+  return name;
+};
--- a/packages/app-extension/src/utils/util.ts
+++ b/packages/app-extension/src/utils/util.ts
@@ -1,6 +1,0 @@
-export const truncateName = (name: string, size: number) => {
-  if (name.length > size) {
-    return name.slice(0, size - 4) + "..." + name.slice(-1);
-  }
-  return name;
-};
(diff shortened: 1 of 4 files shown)


packages/app-extension/src/utils/util.ts
```diff
@@ -0,0 +1,6 @@
+export const truncateName = (name: string, size: number) => {
```
tomlinton, line 1: I don't think we need to introduce a new file for this, we can add it to `@coral-xyz/common` alongside `walletAddressDisplay`. The functions are also largely the same so perhaps that could be refactored.

packages/chat-sdk/src/components/Attachment.tsx
```diff
@@ -5,7 +5,7 @@ import AttachFileIcon from "@mui/icons-material/AttachFile";
 import GifIcon from "@mui/icons-material/Gif";
 import { IconButton } from "@mui/material";
 
-export const Attatchment = ({
+export const Attachment = ({
```
tomlinton, line 8: Unrelated changes here and below

i-m-aditya: This is because of a typo fix [PR](https://github.com/coral-xyz/backpack/pull/2353) which I pushed directly into my master, and it's not yet merged.

aarongillett:
Instead of two full-stops `..` (if I’m interpreting the code correctly…) let’s use a single ellipsis `…`

i-m-aditya:
> Instead of two full-stops `..` (if I’m interpreting the code correctly…) let’s use a single ellipsis `…`
I went with a single ellipsis at first, but after rebasing, I came across a function _userNameDisplay_, and it was using two dots, so I went ahead and updated walletNameDisplay too to follow the same design.

i-m-aditya:
@hkirat Is the PR looking good?

hkirat:
@i-m-aditya build seems to be failing and there seem to be merge conflicts could you fix those

i-m-aditya:
> @i-m-aditya build seems to be failing and there seem to be merge conflicts could you fix those

@hkirat Was getting merge conflicts even after rebasing with the upstream master. Since the changes were small, I created a different PR for this [issue](https://github.com/coral-xyz/backpack/pull/2969).

Issue #2498, opened 2023-02-06
Clicking on a friend request notification should push into the requests screen


armaniferrante:
Clicking on a friend request (not on the buttons) notification should push into the requests screen instead of the contacts screen.

<img width="397" alt="Screen Shot 2023-02-06 at 10 17 34 AM" src="https://user-images.githubusercontent.com/6990215/217025535-5ca11316-afe5-42f5-b56e-ae4858518d08.png">

khalatevarun:
@armaniferrante can you take a look at the [PR](https://github.com/coral-xyz/backpack/pull/2559). I tried fixing the issue.

aarongillett:
I think selecting a friend request and confirmation (e.g. the top two cards here) should both push to that person’s profile. This is the pattern that Twitter and Instagram both use.
<img width="487" alt="image" src="https://user-images.githubusercontent.com/4157473/225007427-2dfbd91f-982c-443e-b737-e4578b78bc15.png">

hackpk:
can i work on this if its still open??

aarongillett:
Thanks @hackpk !

hackpk:
hi @aarongillett, I understand the requirement but can you give me some hints on where to make the changes?

aarongillett:
Sorry @hackpk, I can’t help you with the code (I work on UX design).

hackpk:
Pull request /#3789 is created for the issue. Can anyone please check?

PR #2559, opened 2023-02-08
Fix: redirection on friend request click

khalatevarun:
Fixes Issue #2498

Problem - Clicking on a friend request notification should push into the requests screen #2498

Solution - Replaced `contacts` with `contact-requests` in nav.push function


hkirat:
Were you able to test this change locally @khalatevarun ?
`contact-requests` requires some props to be passed to it when being mounted, don't think you're doing that here
Also this might not work in full screen mode, can you figure out how to achieve that?

PR #2569, opened 2023-02-08
fix webpack-dev-server websocket connection

wentokay:
This should stop these errors from flooding your console when running `yarn start`

Screenshot 2023-02-08 at 14 09 59

It involves some manual steps with creating and installing a self-signed certificate for localhost, here they are for mac OS

```
cd packages/app-extension
brew install mkcert
mkcert localhost
mkcert -install

yarn start
```

If you don't do these steps things will continue to work as before. If this solves an issue for other people we can probably automate it in some way.

These errors exist because the automatically created `webpack-dev-server` certificate is considered [invalid](https://github.com/webpack/webpack-dev-server/issues/2957), at least with the mac OS & openssl versions I am using [office-addin-dev-certs](https://www.npmjs.com/package/office-addin-dev-certs) are also invalid.

---

it also changes the port from 8080 to 9997 (arbitrary number) because that currently conflicts with `backpack-api` and `expo react-native xnft` tooling

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,29 @@ yarn start
 
 Note: In a fresh repo, you should run `yarn build` before `yarn start`.
 
+#### Troubleshooting
+
 _If you run into issues with builds try running `yarn clean` and then start again._
+
+<details>
+  <summary>Seeing `WebSocket connection to 'wss://localhost:9997/ws' failed` error messages in your console?</summary>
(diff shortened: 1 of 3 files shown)


armaniferrante:
Can you add the steps to the readme (and explain how it's optional)?

PR #2574, opened 2023-02-09
plugin messaging fix

wentokay:
fixes issue with xNFTs, after signing a transaction in an xNFT

`post message window not found` would be thrown if developer mode was enabled but the simulator was not being used

`Unknown Origin or channel` would be thrown by the default xNFTs installed in the wallet (prices and explorer)

it also specifies the origin as the event.origin so that it's sent as a reply and not broadcast to all listeners


greenish:
https://github.com/coral-xyz/backpack/pull/2576

wentokay:
closing in favor of https://github.com/coral-xyz/backpack/pull/2576, thanks @greenish!

PR #2576, opened 2023-02-09
Properly tear down plugin event listener when iframes are destroy

greenish:


packages/common/src/plugin.ts
```diff
@@ -191,7 +190,13 @@ export class Plugin {
   //
   public setActiveIframe(iframe: HTMLIFrameElement, xnftUrl: string) {
     this._iframeActive = iframe;
-    this._rpcServer.setWindow(iframe.contentWindow, xnftUrl);
+    // this._rpcServer.handler(this._handleRpc.bind(this));
```
armaniferrante, line 193:
```suggestion
```

PR #2622, opened 2023-02-10
show dropzone xnft in dropzone wallet

wentokay:
### API changes

adds `dropzonePublicKeys: Record<Blockchain, PublicKeyString>` to the `/users/:username` API response

it will only have a  solana key/value now, but this approach should make things a bit easier if we add other blockchain support in future.

I considered adding a dropzone tag to the existing `publicKeys: Array<{ blockchain: Blockchain; publicKey: PublicKeyString }>` array, but decided to keep things simple and make it clear there's only a single dropzone wallet for now

### Extension changes

Added `isDropzoneWallet` in recoil, which is used when loading xNFTs


wentokay:
closing in favor of #2718

Issue #1815, opened 2022-12-12
Add zod validation to all API/websocket endpoints


hkirat:
Right now very easy to send in wrong data.
For eg - https://github.com/coral-xyz/backpack/blob/master/backend/native/backpack-api/src/routes/v1/friends.ts#L40 should be validated to be a `uuid`

wentokay:
it might be worth looking into trpc for this https://trpc.io/docs/procedures#with-zod

gautam2002:
@hkirat I am willing to tackle this. Since you have mentioned to validate all endpoints,  I am thinking of writing a middleware that would handle input/req data validation as per the `baseUrl` and `reqPath `(`req.body` fields mostly vary from path to path). So you could assign me if others aren't working on it

hkirat:
Hi @gautam2002 , that’s great
What’s required here is using Zod to verify input parameters
A middleware sounds good but it would need to be different for every endpoint since all endpoints have different input Params
Let me know if you have any questions

gautam2002:
Thanks for the concern @hkirat. I think I can leverage the `req.Path` provided by express js to identify and load the required zod schema based upon the endpoint and validate against either req object without mutating or extracting what input params are required to validate. 
Here's how:
For instance, when req is made to this:  `/user/add`  middleware would extract the basepath (`user`) and reqPath(`add`) and based on these it would switch the logic of loading schema and data validation. However, this model couldn't be feasible if a new endpoint needed to be added, the dev had to make new adjustments at many places. Therefore (priority) I would try to further improve this middleware model else would write a validator function that takes schema and req object and next(to forward to err). word of suggestion would be appreciated

hkirat:
Sounds good @gautam2002 , looking forward to seeing your implementation

gautam2002:
@hkirat drafted a pull request #2633

PR #2633, opened 2023-02-11
added zod validation api endpoints

Xavek:
### WHY?

Added zod validation to most API endpoints. See for ref: #1815 

### HOW?

Earlier thought was to write middleware for handling req[data] validation, based upon the requested path/route the schema logic would be switched and validated/parsed against the req object without mutating it. However due to the high entropy (every endpoint consumes unique data), extending, omit/pick on base zod schema was helpless. Furthermore handling the edge case like the same route of “/” with different HTTP methods contributed to the long switch cases. Plus adding a new route required making adjustments at many places.

Therefore. I wrote a middleware wrapper (sort of), which takes zod schema/types as a param and validates/parses them against a req object.If succeeded next would be call else next(err)[for global middleware] 
```
router.get(
 "/",
 bodyValidator(
   z.object({
     query: z.object({
       usernamePrefix: z.string(),
     }),
     id: z.string().uuid(),
 })
 ),
 extractUserId,
 async (req, res) => {
   // @ts-ignore
   const usernamePrefix: string = req.query.usernamePrefix;
   // @ts-ignore
   const uuid = req.id as string;
```
The above code depicts how to write validation schema if needed to validate `req.query.usernamePrefix` and `req.id`. 
### Anything
If the author seems to be okay with the approach can further strengthen the sanitation and validation rules by forcing custom types for `publicKey, etc`. Also incorporating OWASP validation rules.

### DIFF
--- a/backend/native/backpack-api/src/routes/v1/chats.ts
+++ b/backend/native/backpack-api/src/routes/v1/chats.ts
@@ -4,6 +4,7 @@ import type {
   SubscriptionType,
 } from "@coral-xyz/common";
 import express from "express";
+import { z } from "zod";
 
 import { ensureHasRoomAccess, extractUserId } from "../../auth/middleware";
 import {
@@ -15,11 +16,27 @@ import {
   updateLastReadGroup,
   updateLastReadIndividual,
 } from "../../db/friendships";
-
+import { bodyValidator } from "../../validation/reqValidationMiddleware";
 const router = express.Router();
 
 router.post(
   "/lastRead",
+  bodyValidator(
+    z.object({
+      body: z.object({
+        client_generated_uuid: z.string().uuid(),
+      }),
+      id: z.string().uuid(),
+      query: z.object({
+        room: z.string(),
+        type: z.enum(["collection", "individual"]),
+      }),
+      roomMetadata: z.object({
+        user1: z.string(),
+        user2: z.string(),
+      }),
+    })
+  ),
   extractUserId,
   ensureHasRoomAccess,
   async (req, res) => {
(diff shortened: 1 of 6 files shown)


backend/native/backpack-api/src/routes/v1/chats.ts
```diff
@@ -48,48 +65,83 @@ router.post(
...
+        room: z.string(),
+      }),
+    })
+  ),
+  extractUserId,
+  ensureHasRoomAccess,
+  async (req, res) => {
+    //TODO: make this secure, there is a path to cancel but the UI shows the txn as redeemed.
+    const room = req.query.room;
+    const messageId = req.body.messageId;
```
hkirat, line 87: At this point since we've validated the schema using zod, is there a way we can have `req.query` or `req.body` be typed better?
This comment applies across all the routes

backend/native/backpack-api/src/routes/v1/chats.ts
```diff
@@ -15,11 +16,27 @@ import {
...
 const router = express.Router();
 
 router.post(
   "/lastRead",
+  bodyValidator(
+    z.object({
+      body: z.object({
+        client_generated_uuid: z.string().uuid(),
+      }),
+      id: z.string().uuid(),
```
hkirat, line 29: This might not work since `id` is added as a field in the `extractUserId` middleware
I don't think there is need to validate these either. Only need to validate user provided arguments

hkirat:
Overall looks like the right approach, added a few comments

backend/native/backpack-api/src/routes/v1/chats.ts
gautam2002: I tried to sync `req.body` with zodSchema, intention was to make something like trpc server router, and did extend `Express.router` interface but didn’t get lucky. Now most obvious to have `req.body` typed would be to define Zod schema first, extract out types from inferring, and passing down to `request<T>`. Another option would be to make a generic type like `[key:string]:string` but this wouldn’t be compatible with all the routes. Let me know if you have any ideas or resources, or examples of how should i approach this.

Furthermore, if this thing isn’t necessary should I start adding more validation rules and address another comment about omitting `req.id` checks?

Issue #2677, opened 2023-02-13
Send visual and ux updates


trucy:
Figma ([Editor](https://www.figma.com/file/HNF2HJZYq6dUzCWxSmH2Oe/Backpack-_-Questlog?node-id=18919%3A149106&t=HQ6InyXbsJVnldCA-1))
Figma ([Prototype](https://www.figma.com/proto/HNF2HJZYq6dUzCWxSmH2Oe/Backpack-_-Questlog?page-id=18899%3A148945&node-id=18919%3A149106&viewport=879%2C-977%2C1&scaling=min-zoom&starting-point-node-id=18919%3A149106))

PR #2686, opened 2023-02-14
Added new designs to send flow

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/2677


hkirat:
Merging a checkpoint, will follow up with more

PR #2740, opened 2023-02-16
Update Dockerfile

AtulRajput01:


hkirat:
@AtulRajput01 could you explain the reasoning behind this change?
Would it be faster/any other benefit?
I like the way it looks originally, easier to read.

hkirat:
Is it to minimize docker layers?

AtulRajput01:
> Is it to minimize docker layers?

yes , i am Reduce number of layers in docker development file. it will increase the efficiency of the container.

AtulRajput01:
> @AtulRajput01 could you explain the reasoning behind this change? Would it be faster/any other benefit? I like the way it looks originally, easier to read.

@hkirat I'm very concerned about the number of layers in container images that I always not build locally and push to a container repository, but no problem to simplify.

As a way to make it clearer, I think I'll move the wget test earlier so that I don't need to ask hard questions about the order of evaluation of  &&

PR #2822, opened 2023-02-19
Update Dockerfile

AtulRajput01:
Reduce number of layers in docker development file


tomlinton:
Please consolidate into https://github.com/coral-xyz/backpack/pull/2825

PR #2826, opened 2023-02-19
Fixed dockerfile for backpack api

hkirat:


hkirat:
Follow up from https://github.com/coral-xyz/backpack/pull/2823 which broke the dockerfile

AtulRajput01:
i would like to work on it , please assign this to me @hkirat sir

AtulRajput01:
> Follow up from #2823 which broke the dockerfile

@hkirat i'm sorry i think that's my mistek , i will fix it

PR #2827, opened 2023-02-19
Fixed broken Dockerfile #2826

AtulRajput01:


tomlinton:
Done in https://github.com/coral-xyz/backpack/pull/2826

PR #2855, opened 2023-02-20
feat: use Intl.NumberFormat notation to format large nums

### LINKED ISSUE #2811: prevent wrap on negative/positive symbol in token symbol
Negative/positive symbol should never be on a line on its own.

Screenshot 2023-02-19 at 9 53 14 AM

jim4067:
Resolves #2811 

Screenshot from 2023-02-20 14-39-21


hkirat:
Hi @jim4067, If I'm not mistaken we lose precision with this change (1003 resolves to 1k, 1050 resolves to 1.1k). 
I think that could lead to confusion especially because these are token balances

jim4067:
@hkirat thanks for pointing that out.

Issue #2935, opened 2023-02-23
Make hotreloading in the app-extension smooth as butter


hkirat:
Hot reloading is a mess currently. 
Have to compile a lot of modules independently if you've touched more than the app-extension.

wentokay:
💯  I think a big step in that direction would be to get rid of all the `@coral-xyz/*` packages and do relative imports 🧈

armaniferrante:
can this be closed?

Issue #2893, opened 2023-02-22
remove zustand dependency from @coral-xyz/common


tomlinton:
We need to remove this dependency because it also requires `react` and thus makes `backpack-api` dependent on `react` via `@coral-xyz/common`

This should be reverted https://github.com/coral-xyz/backpack/pull/2892

tomlinton:
Also this

<img width="814" alt="Screenshot 2023-02-22 at 11 12 50 PM" src="https://user-images.githubusercontent.com/489202/220589996-2a36d42d-d8e4-45de-ac69-cfff0d0e921b.png">

hkirat:
Yeah common shouldnt contain any server/frontend library. 
server libraries should reside in `common-backend`
frontend libraries should reside in `common-react`
If frontend and only used by mobile - `common-mobile`

hkirat:
Just to point out we reverted hot reloading a while back and this might come after that happens

armaniferrante:
Can re-open if it comes back.

PR #2943, opened 2023-02-23
Hot reloading, new modules

hkirat:
Related https://github.com/coral-xyz/backpack/issues/2935

https://user-images.githubusercontent.com/8079861/221039053-46564c2b-cb05-46f2-89a4-2f77d41c679e.mov

The PR 
1. Creates a new `background-injection` module. This is done so that hot reloading in `app-extension` doesnt affect the background module. This is done because the background goes down anytime it hot reloads, causing the app to stop working and the user has to go to the extensions tab and refresh it. 
2. Creates a common-mobile module that `mobile` can import for stuff that doesnt belong to `common`. This is to avoid the `common` module which is imported by `app-extension` to have any mobile dependencies (like `expo-secure-store`)

With these two changes, `app-extension` hot reloads like butter (check video)
Only thing that doesnt hot reload is `background`. So if you ever make changes in the service worker, you still have to rebuild it. We should fix this , but this module isnt touched 95% of the time so not a top priority atm.


hkirat:
cc @peterpme I've touched mobile code, would be great if we could check if I didnt break anything.

packages/app-extension/webpack.config.js
```diff
@@ -217,6 +216,11 @@ const options = {
           to: "injected.js",
           force: true,
         },
+        {
```
hkirat, line 219: Background copied over to avoid hot reloading like before

packages/common/src/browser/index.ts
```diff
@@ -1,8 +1,3 @@
-import { startMobileIfNeeded } from "./mobile";
-
 export * from "./common";
 export * from "./extension";
-export * from "./mobile";
 export * from "./uiActionRequestManager";
-
-startMobileIfNeeded();
```
hkirat, line 8: Moved to `common-mobile`

hkirat:
Will wait for todays release cut before merging this in

hkirat:
Just checking in is this harmless @peterpme ?
Screenshot 2023-02-24 at 3 52 22 AM

tomlinton:
Any chance this helps with https://github.com/coral-xyz/backpack/issues/2893? Could that live in `common-mobile`?

hkirat:
yeah probably would fix this, will add that

PR #2954, opened 2023-02-24
removed limit on friendship requests

hkirat:


hkirat:
Causing messages to dissaper. We need a longer term fix that paginates the inbox

Issue #2760, opened 2023-02-16
Drawer polish (light mode only)


trucy:
- update opacity of drawer overlay background from 0.8 to 0.2
- update border color around close X for drawers to #F0F0F2

left: current, right: proposed

<img width="678" alt="Screenshot 2023-02-16 at 3 48 47 PM" src="https://user-images.githubusercontent.com/321395/219494764-d3097d8c-a75d-44ef-b92e-2dc391086043.png">

venkata-krishnas:
Hi, I am a beginner and I would love to work on this issue. Can you please help me on how I can find the file where I can modify the code. Thanks a lot.

kartik-0908:
Hey, i would love to contribute to fix this issue, assign it to me if it's still not resolved

trucy:
@kartik-0908 assigned!

kartik-0908:
Hey @trucy  created a PR for this issue changing the opacity and background color as suggested

PR #2976, opened 2023-02-24
Drawer styles updates

Kritik-J:
Fixes Issue - https://github.com/coral-xyz/backpack/issues/2760

Problem - Drawer polish (light mode only) #2760

Solution - Changed backdropProps style opacity from 0.8 to 0.2 and approveTransactionCloseBackground from #C2C4CC to #F0F0F2


packages/app-extension/src/components/common/ApproveTransactionDrawer.tsx
```diff
@@ -44,7 +44,7 @@ export const ApproveTransactionDrawer: React.FC<{
       setOpenDrawer={setOpenDrawer}
       backdropProps={{
         style: {
-          opacity: 0.8,
+          opacity: 0.2,
```
armaniferrante, line 47: there's a lot more places where thsi is used

packages/themes/src/index.tsx
```diff
@@ -329,7 +329,7 @@ const LIGHT_COLORS: CustomColors = {
   swapTokensButton: "#FFFFFF",
   icon: "#8F929E",
   approveTransactionTableBackground: LIGHT_BACKGROUND_COLOR_1,
-  approveTransactionCloseBackground: "#C2C4CC",
+  approveTransactionCloseBackground: "#F0F0F2",
```
armaniferrante, line 332: The original issue says to change the border. Does this do that?

armaniferrante:
Please share images of these changes

packages/themes/src/index.tsx
Kritik-J: Yes, as I did not find any border property here. Instead, I discovered that the close button in light mode has a backgroundColor of "#FFFFFF" and a width and height of 38 px, which is wrapped inside a div with a backgroundColor of #C2C4CC and a width and height of 44 px.

packages/app-extension/src/components/common/ApproveTransactionDrawer.tsx
Kritik-J: Thanks for pointing this out; I will fix others as well.

packages/themes/src/index.tsx
Kritik-J: This is the output -

drawer

Kritik-J:
@armaniferrante These are some images of the changes

drawer-1

drawer-2

drawer-3

drawer-4

drawer-5

drawer-6

PR #3044, opened 2023-02-28
Allow all uuids as invite codes

hkirat:


wentokay:
related cf issue https://www.cloudflarestatus.com/incidents/cx215c2tx7yw

PR #3066, opened 2023-03-01
autofocus message input

yash-w3:
Yooo new heree, I am a Open source dev. Created web3 gh account for now. 

**Changes**

1. Autofocus the message input, added the flag to prevent autofocus in case its needed
2. Fixed lint issues & minor refactor

Fixes: #3032

### DIFF
--- a/packages/chat-sdk/src/components/messageInput/MessageInput.tsx
+++ b/packages/chat-sdk/src/components/messageInput/MessageInput.tsx
@@ -1,4 +1,4 @@
-import { useContext, useMemo } from "react";
+import { useContext, useEffect, useMemo } from "react";
 import { RichMentionsContext, RichMentionsInput } from "react-rich-mentions";
 import { useUsersMetadata } from "@coral-xyz/react-common";
 import { styles, useCustomTheme } from "@coral-xyz/themes";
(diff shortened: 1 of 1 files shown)


packages/chat-sdk/src/components/messageInput/MessageInput.tsx
```diff
@@ -14,21 +14,34 @@ const useStyles = styles(() => ({
...
 }) {
-  const defaultValue = "";
   const classes = useStyles();
   const theme = useCustomTheme();
   const { type, remoteUsername, activeReply } = useChatContext();
   const { activeSearch } = useContext(RichMentionsContext);
 
+  useEffect(() => {
+    if (autoFocus) {
+      const messageElement = document.getElementById(messageInputElementId);
```
hkirat, line 33: Can you use a ref here instead of `document.getElementById` ?

hkirat (changes requested):

packages/chat-sdk/src/components/messageInput/MessageInput.tsx
yash-w3: No, actually we can't use it here, if you see the library code they are not using `forwardRef` to pass `ref` to the underlying element. I did give it a try but turned out I was getting `undefined` since its ref to function component which returns undefined or null based on default value you pass

hkirat (approved):

packages/chat-sdk/src/components/messageInput/MessageInput.tsx
hkirat: Note- For now merged it in but might be good to clean it up later on

PR #3073, opened 2023-03-01
Fixes nft breaking issuye

hkirat:


packages/common/src/utils.ts
```diff
@@ -6,7 +6,7 @@ import { v1 } from "uuid";
 import { IMAGE_PROXY_URL } from "./constants";
 
 export function toTitleCase(str: string) {
-  return str.slice(0, 1).toUpperCase() + str.toLowerCase().slice(1);
+  return str?.slice(0, 1)?.toUpperCase() + str?.toLowerCase()?.slice(1);
```
peterpme, line 9: the type says`str: string` but then you're adding a question mark, so should that type be an optional string?

hkirat: yeah the type should be an optional string correct . Only happens when the metadata of an NFT is messed up and doesnt return a string

PR #3088, opened 2023-03-02
update swap ui to handle fees when they're enabled

wentokay:
Without fees | With fees
-------------|-----------
Screenshot 2023-03-01 at 17 19 20 |  Screenshot 2023-03-01 at 17 19 00

### DIFF
--- a/packages/recoil/src/context/Swap.tsx
+++ b/packages/recoil/src/context/Swap.tsx
@@ -1,4 +1,3 @@
-import React, { useContext, useEffect, useRef, useState } from "react";
 import {
   Blockchain,
   confirmTransaction,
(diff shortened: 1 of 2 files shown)


packages/recoil/src/context/Swap.tsx
```diff
@@ -154,10 +174,40 @@ export function SwapProvider({
...
+              .mulUnsafe(
+                FixedNumber.from(100).divUnsafe(
+                  FixedNumber.fromString(swapFee.pct.toString())
+                )
+              )
+              .ceiling()
+              .toString()
+              .split(".")[0]
+          );
+        } else {
```
wentokay, line 197: There's probably a better way to do this but the predicted output amounts were not accurate when fees were included. I tried various approaches including removing fees from the predicted total received (adding fees to the total ended up being closer to the actual result) etc.

This seems to work, but I'll run it by the jupiter team to confirm it's not making any false assumptions.

It doesn't affect affect existing swaps without fees, they still use `route.outAmount` as they have been doing until now

PR #3099, opened 2023-03-02
Implement the single mnemonic text input box

staticGuru:
Feature implementations

image


staticGuru:
@armaniferrante It is very nice to implement the feature
When do you merge my changes?

tomlinton:
@staticGuru appreciate the PR but I don't believe this is a feature we want. Users can paste a mnemonic and it just splits it up into the text boxes so I don't see any advantage here. 

The Issues tab is a good place to start if you are looking for features to implement.

PR #3139, opened 2023-03-03
Fixes squished PFPs and Desktop NFT Details View

irffanasiff:
Fixes
1) Nfts with different dimensions did not look good
Screenshot 2023-03-03 at 4 32 05 PM
Screenshot 2023-03-03 at 4 32 35 PM
Screenshot 2023-03-03 at 4 33 55 PM
2) Desktop NFT view
NFTs looked enlarged in desktop
Screenshot 2023-03-03 at 4 41 23 PM

Screenshot 2023-03-03 at 4 42 03 PM


packages/react-common/src/components/base/ProxyImage.tsx
```diff
@@ -55,6 +55,7 @@ export const ProxyImage = React.memo(function ProxyImage({
         style={{
           ...(imgProps.style ?? {}),
           ...visuallyHidden,
+          objectFit: "cover",
```
hkirat, line 58: `ProxyImage` is used throughout the project.
Can you confirm if this change doesn't cause a regression elsewhere?

irffanasiff: The default value of object-fit is fill, which stretches the image to fill the entire container without maintaining its aspect ratio. This can cause the image to become distorted if its aspect ratio is different from the container's aspect ratio.

To avoid distortion, it's a good idea to use object-fit: cover or object-fit: contain. object-fit: cover maintains the aspect ratio of the image while covering the entire container.

Using object-fit: cover on all images will ensures that the aspect ratio of the image is maintained

irffanasiff: Screenshot 2023-03-03 at 10 46 24 PM
Screenshot 2023-03-03 at 10 46 45 PM
Screenshot 2023-03-03 at 10 47 19 PM
all these are distorted except the last one but it is not a circle, issue with current implementation

PR #3153, opened 2023-03-03
fix: update profile screen

### LINKED ISSUE #2550: Improve contact profile page
1. This user is not yet a contact, but there’s a lock icon that makes them seem safe. We should remove it.
Screenshot 2023-01-12 at 16 06 46

2. The verified icon shouldn’t look and feel like a button.
Before:
image
After:
- Icon moved to after “Connected” text
- Size reduced to 0.8em (or similar)
- Space between text reduced.
image

3. More info icon + menu should also display on contact page.
This menu: 
image

rt-web3:
Closes #2550 
Screenshot 2023-03-04 at 12 11 52 AM
Screenshot 2023-03-04 at 12 12 17 AM


rt-web3:
@hkirat not sure if this is urgent but would love feedback on this

aarongillett:
Considering we’re now using the verified icon for Backpack staff, we should update the mutual friend icon to the filled version of Material Icon: `sentiment_satified`

PR #3182, opened 2023-03-05
Collectibles update

ph101pp:
Screen Shot 2023-03-05 at 4 03 29 PM
Screen Shot 2023-03-05 at 4 04 59 PM


armaniferrante:
xnft hover state doesn't change on hover

Issue #3188, opened 2023-03-06
Make call to get friends faster


hkirat:
We now have too many public keys and this call is very slow - https://github.com/coral-xyz/backpack/blob/master/backend/native/backpack-api/src/db/users.ts#L50

We should do the public keys call and user call separately and merge them later on

Issue #3187, opened 2023-03-06
messages: dms not showing


armaniferrante:
<img width="443" alt="Screen Shot 2023-03-06 at 12 06 24 PM" src="https://user-images.githubusercontent.com/6990215/223193883-c405a93f-2837-485a-be53-d9837090be51.png">

hkirat:
Most probably this is caused by https://github.com/coral-xyz/backpack/issues/3188

PR #3189, opened 2023-03-06
Made DB request faster

hkirat:
Closes https://github.com/coral-xyz/backpack/issues/3188
Potentially related https://github.com/coral-xyz/backpack/issues/3187


hkirat:
Will optimise further in follow up by merging all DB requests

Issue #3193, opened 2023-03-06
Remove reply and delete icons for deleted messages


trucy:
You shouldn't be able to reply to or delete a previously deleted message

(remove the icons on hover)

<img width="373" alt="Screenshot 2023-03-06 at 1 51 46 PM" src="https://user-images.githubusercontent.com/321395/223215876-f9aab3f1-70e3-486a-a0a4-f07f6c382230.png">

PR #3205, opened 2023-03-07
Removed reply and delete icons when hover deleted messages

Kritik-J:
Closes: Remove reply and delete icons for deleted messages #3193

### DIFF
--- a/packages/chat-sdk/src/components/Message.tsx
+++ b/packages/chat-sdk/src/components/Message.tsx
@@ -333,51 +333,53 @@ export const MessageLine = (props) => {
                     )}
                   </p>
                 </div>
-                <div>
-                  {props.messageKind === "text" ? (
-                    <div
-                      style={{ display: "flex" }}
-                      className={classes.hoverChild}
-                    >
+                {props.deleted ? null : (
+                  <div>
+                    {props.messageKind === "text" ? (
                       <div
-                        style={{
-                          marginLeft: 10,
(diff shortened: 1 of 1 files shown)


hkirat (changes requested):

packages/chat-sdk/src/components/Message.tsx
```diff
@@ -334,7 +334,7 @@ export const MessageLine = (props) => {
                   </p>
                 </div>
                 <div>
-                  {props.messageKind === "text" ? (
+                  {props.messageKind === "text" && props.deleted ? (
```
hkirat, line 337: should it be `!props.deleted` ?

hkirat (approved):

hkirat (approved):

packages/chat-sdk/src/components/Message.tsx
```diff
@@ -334,7 +334,7 @@ export const MessageLine = (props) => {
                   </p>
                 </div>
                 <div>
```
hkirat, line 336: We should add the same check here https://github.com/coral-xyz/backpack/pull/3205/files#diff-c4599c5c66c69aaef29a6ed1bdc5ac4b9daacbe84205d939a8da0c080c11bde0L371
(search for `DeleteIconInternal` through the file')

hkirat (changes requested):
Added one more comment

PR #3214, opened 2023-03-07
close any existing backpack windows when opening a new one

wentokay:
Closes other instances of backpack upon opening, for both extension and popout mode.

When a user has multiple backpack windows open it can lead to unpredictable behaviors.

https://user-images.githubusercontent.com/101902546/223525604-5350f328-c075-4ce1-b3b0-32be6d728e50.mov

I don't yet know what the implications of this would be, it needs more testing before being merged.

There could be unforeseen issues with the autolock timer not being restarted for instance

### DIFF
--- a/packages/app-extension/src/index.tsx
+++ b/packages/app-extension/src/index.tsx
@@ -1,4 +1,4 @@
-import React, { lazy, Suspense } from "react";
+import { lazy, Suspense } from "react";
 import { createRoot } from "react-dom/client";
 import { BACKPACK_FEATURE_POP_MODE, openPopupWindow } from "@coral-xyz/common";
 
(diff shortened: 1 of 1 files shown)


packages/app-extension/src/index.tsx
```diff
@@ -7,6 +7,17 @@ import "./index.css";
 const App = lazy(() => import("./app/App"));
 const LedgerIframe = lazy(() => import("./components/LedgerIframe"));
 
+// Tell all existing extension instances that this instance now exists
```
armaniferrante, line 10:
```suggestion
// Tell all existing extension instances that this instance now exists.
// This block ensures a single extension window is open at any given time.
```

packages/app-extension/src/index.tsx
```diff
@@ -7,6 +7,18 @@ import "./index.css";
...
 const LedgerIframe = lazy(() => import("./components/LedgerIframe"));
 
+// Tell all existing extension instances that this instance now exists.
+// This block ensures a single extension window is open at any given time.
+chrome.runtime
+  .sendMessage("new-instance-was-opened")
+  .then(() => {
+    // Close all existing extension instances so only the newest is running
+    chrome.runtime.onMessage.addListener((msg) => {
+      if (msg === "new-instance-was-opened") window.close();
```
armaniferrante, line 17: Can you add a check with `if(!isValidEventOrigin(msg)) return;` here. I don't think it's technically needed but would be good to have. (can import from @coral-xyz/common

armaniferrante: https://github.com/coral-xyz/backpack/blob/6c59f0325ebbf2421f48ab1a5954a2ddd4466b4f/packages/common/src/index.ts#L113

wentokay: now done in b3b7d4f0444adb66966d3a548f839ad1fe6935b4

PR #3221, opened 2023-03-07
fix ui issue when sending to .sol domains

wentokay:
fixes a bug where it looked like the public key you are sending to was the .sol domain

#### sending to `backpack.sol`

Screenshot 2023-03-07 at 15 11 34

before | after
-----|-----
Screenshot 2023-03-07 at 15 12 15|Screenshot 2023-03-07 at 15 11 38


wentokay:
tested with .eth domains too

PR #3235, opened 2023-03-08
tamagui across extension and mobile

peterpme:
- wip
- wip


packages/app-extension/webpack.config.js
```diff
@@ -69,6 +70,14 @@ const swcLoaderConfiguration = {
   },
 };
 
+const tamaguiLoaderConfiguration = {
```
armaniferrante, line 73: Q: is this needed if all the tamgui usage is contained to a shared package used by the extension and mobile app?

PR #3249, opened 2023-03-09
fix size of popout for windows users

### LINKED ISSUE #547: Ensure popping window doesn't cut the bottom bar
image

wentokay:
fixes #547

## before

popout on right of regular extension window, it's slightly shorter and a lot thinner

Screenshot 2023-03-09 at 08 26 57

## after

sizes now match

Screenshot 2023-03-09 at 08 26 17

### DIFF
--- a/packages/common/src/browser/extension.ts
+++ b/packages/common/src/browser/extension.ts
@@ -54,56 +54,9 @@ export class BrowserRuntimeExtension {
     // that those promises can properly resolve with the right state,
     // i.e. user denied the request.
     //
-    UiActionRequestManager.cancelAllRequests();
-
-    return new Promise(async (resolve, reject) => {
-      // try to reuse existing popup window
-      try {
-        const popupWindowId: number | undefined =
-          await BrowserRuntimeCommon.getLocalStorage("popupWindowId");
-        if (popupWindowId) {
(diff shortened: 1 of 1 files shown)


packages/common/src/browser/extension.ts
```diff
@@ -54,56 +54,9 @@ export class BrowserRuntimeExtension {
...
-          );
-          resolve(newPopupWindow);
-        }
-        const error = BrowserRuntimeCommon.checkForError();
-        return reject(error);
-      } catch (e) {
-        const error = BrowserRuntimeCommon.checkForError();
-        return reject(error);
-      }
-    });
```
wentokay, line 106: @ph101pp I removed some of these lines added in #1941 because #3214 will now close any existing popout or extension windows when you click the backpack icon

The previous behavior was that you could have the popout and a regular extension window at the same time

https://user-images.githubusercontent.com/101902546/224091580-29737f79-93ff-4f42-a70c-c8ed8a941f25.mov

but now this is impossible because it will close the popout when you reopen the extension, so you will only ever have 1 backpack window open

I might have misunderstood something though so lmk if you think I should keep this code in. Thanks!

ph101pp: Yea that makes sense no need for this then

Issue #3246, opened 2023-03-09
cannot nfts send to ANS or SNS domains 


crypt0miester:
We have experienced tickets that users could not send NFTs to ANS or SNS domains.

I went through the code and found that the [useIsValidAddress](https://github.com/coral-xyz/backpack/blob/dcf709a41783c28b40b79cb4b54fb23c743a6db1/packages/app-extension/src/components/Unlocked/Nfts/Detail.tsx#LL409C11-L409C25) while sending the nft does not take in normalizedAddress like in how tokens are being sent in [here](https://github.com/coral-xyz/backpack/blob/dcf709a41783c28b40b79cb4b54fb23c743a6db1/packages/app-extension/src/components/Unlocked/Balances/TokensWidget/Send.tsx#LL234) 

would be amazing if it could be added. from where I see it we just need to add destinationAddesss from the hook.

thanks lads and have an amazing day.

wentokay:
we could also enable sending NFTs to other `@usernames`

crypt0miester:
> we could also enable sending NFTs to other `@usernames`

certainly.

armaniferrante:
That's being handled here https://github.com/coral-xyz/backpack/issues/2733

hackpk:
Can i work on this if its not fixed yet?

hackpk:
@crypt0miester what changes do you need ??

hackpk:
that great @crypt0miester, Is there any other similar issue you can recommend for me to contribute??

PR #3252, opened 2023-03-09
Adds contacts logic to nfts

hkirat:
closes https://github.com/coral-xyz/backpack/issues/2733
https://user-images.githubusercontent.com/8079861/224125569-d78538eb-ea77-4cc3-bba0-399928409750.mov


wentokay:
if this handles ANS, SNS and .eth domains as well it would also cover #3246

hkirat:
This doesn't yet handle domains, but should be easy to add

armaniferrante:
anything blocking this from merging?

Issue #3276, opened 2023-03-10
window immediately closes after connecting wallet


armaniferrante:
This was introduced post v0.5.2. Release blocker.

PR #3280, opened 2023-03-10
reimplement #3249

wentokay:
reimplements correct popout size in windows changes (#3249) with fix for #3276 included

### DIFF
--- a/packages/common/src/browser/extension.ts
+++ b/packages/common/src/browser/extension.ts
@@ -54,56 +54,9 @@ export class BrowserRuntimeExtension {
     // that those promises can properly resolve with the right state,
     // i.e. user denied the request.
     //
-    UiActionRequestManager.cancelAllRequests();
-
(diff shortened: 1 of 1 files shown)


packages/common/src/browser/extension.ts
```diff
@@ -216,40 +169,25 @@ export async function openApproveMessagePopupWindow(
...
+    type: "popup",
+    width: EXTENSION_WIDTH + EXTRA_WIDTH,
+    height: EXTENSION_HEIGHT + EXTRA_HEIGHT,
+    top: lastWindow.top,
+    left:
+      (lastWindow.left ?? 0) +
+      ((lastWindow.width ?? 0) - EXTENSION_WIDTH - EXTRA_WIDTH),
+    focused: true,
   });
+  return popupWindow;
```
wentokay, line 190: this was effectively returning `_window` before

PR #3291, opened 2023-03-11
fix: UI hard to navigate for screenreader users #3240

### LINKED ISSUE #3240: UI hard to navigate for screenreader users
Dear Backpack team,
with this issue I as a blind user would like to report a few obstacles in the UI.
* During the setup, you have to choose networks. Checkboxes would be ideal for that, since you toggle them on and off, exactly as needed here. However, buttons are used instead, making it impossible for me to see what is toggled on or off.
* Same applies in some other phases in the setup as well.
* In the main UI, most of the buttons aren't labelled at all. Some have a description below them, but this is not clear when navigating the UI with tab, since interactive HTML elements should have associated labels.
* In some points, hover-elements are used. Please replace them by their semantic HTML equivalents.
* The button to copy the public address should get the label "copied" only after it was pressed, however I see this label right when entering the UI. This is not what it meant to be, because it isn't copied before I click the button.
* As a summary: Please keep to HTML standards, and - but only - where necessary aria-attributes can be used if native HTML isn't suitable for the given use case.
I'm looking forward to seeing an accessible cryptocurrency wallet in the future!

staticGuru:
Closes #3240


packages/app-extension/src/components/common/WalletList.tsx
```diff
@@ -808,19 +816,33 @@ function CopyButton({
...
-        e.stopPropagation();
-        setIsCopying(true);
-        setTimeout(() => setIsCopying(false), 1000);
-        onClick();
-      }}
-      inverted={inverted}
-    >
-      <>{isCopying ? "Copied!" : "Copy"}</>
-    </WalletListButtonBase>
+    <WithCopyTooltip tooltipOpen={tooltipOpen}>
```
armaniferrante, line 822: Can we remove this change and consider it separately?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
```diff
@@ -81,7 +81,7 @@ export const BlockchainSelector = ({
                   selectedBlockchains.includes(Blockchain.SOLANA) ? (
                     <CheckBadge />
                   ) : (
-                    ""
+                    <UnCheckBadge />
```
armaniferrante, line 84: Can we remove this change and consider it separately?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
```diff
@@ -67,7 +67,7 @@ export const BlockchainSelector = ({
                   selectedBlockchains.includes(Blockchain.ETHEREUM) ? (
                     <CheckBadge />
                   ) : (
-                    ""
+                    <UnCheckBadge />
```
armaniferrante, line 70: Can we remove this change and consider it separately?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
```diff
@@ -145,8 +145,8 @@ function CheckBadge() {
       style={{
         display: "inline-block",
         position: "relative",
-        top: "4px",
-        left: "5px",
+        top: 0,
```
armaniferrante, line 148: Can we remove this change and consider it separately?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
staticGuru: @armaniferrante what do you mean here?
Do we need to create a separate PR for each type of change?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
staticGuru: @armaniferrante what do you mean here?
Do we need to create a separate PR for each type of change?

packages/app-extension/src/components/Onboarding/pages/BlockchainSelector.tsx
armaniferrante: Not for each change just for the comments here.

PR #3311, opened 2023-03-12
adds sentry to backpack-api and app-mobile

peterpme:
adds sentry perf / error monitoring

i need help with:
- adding environment variables (SENTRY_DSN)
- knowing which fields to filter out
- knowing which services to add as projects


packages/app-mobile/src/App.tsx
```diff
@@ -18,6 +18,7 @@ import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
 import { SafeAreaProvider } from "react-native-safe-area-context";
 import { WebView } from "react-native-webview";
 import { RecoilRoot } from "recoil";
+import * as Sentry from "sentry-expo";
```
hkirat, line 21: Should we avoid this until we audio sentry-expo? Like adding it to the app-extension/app-mobile?

peterpme: sentry is already hooked up!

PR #3330, opened 2023-03-13
Mobile messaging base, adds indexed db to mobile

hkirat:
The PR 

1. Moves a bunch of components from `react-common` to `tamagui` so they can be used across app-extension and mobile
2. Makes indexed db work on mobile
3. Creates a simple component that renders all messages, lets you select a message room and chat inside it.
More context in video

https://user-images.githubusercontent.com/8079861/224817607-0e9e4b08-765d-4477-8bab-10106c048f03.mp4


packages/chat-sdk/src/components/Message.tsx
```diff
@@ -1044,9 +1046,7 @@ function MessageLeft(props) {
                 parent_username: `@${props.username}`,
                 parent_message_author_uuid: props.userId,
               });
-              document
-                .getElementById(chatMessageInputId)
-                ?.focus();
+              document.getElementById(chatMessageInputId)?.focus();
```
peterpme, line 1049: document.getElementById won't work on mobile but i think you know this

packages/db/src/db/getIndexDb.native.ts
```diff
@@ -0,0 +1,4 @@
+export const getIndexDb = () => ({
```
peterpme, line 1: can you add a comment above here so we know what it's for?

packages/db/src/db/index.ts
```diff
@@ -8,6 +8,8 @@ import type {} from "@coral-xyz/common/src/messages/fromServer";
 import type { Table } from "dexie";
 import Dexie from "dexie";
 
+import { getIndexDb } from "./getIndexDb";
```
peterpme, line 11: you might need to place a `@ts-expect-error` above here (or somewhere, i see the build is failing)

packages/tamagui-core/src/index.tsx
```diff
@@ -1,5 +1,6 @@
+export * from "./chat";
 export * from "./components";
-export { useCustomTheme } from "./hooks";
+export * from "./hooks";
```
peterpme, line 3: tamagui should only have components in it, the other hooks probably belong in recoil

packages/chat-sdk/src/components/Message.tsx
hkirat: `chat-sdk` isn't used by `app-mobile` , only imported by app-extension

packages/db/src/db/index.ts
hkirat: The two files I added are `getIndexDb.ts` and `getIndexedDb.native.ts` so this should be fine

packages/tamagui-core/src/index.tsx
hkirat: Moved all to `hooks/chat`

hkirat:
Checked in with @peterpme , merging this in!

PR #3379, opened 2023-03-16
upgrade @onsol/tldparser to latest

crypt0miester:
We upgraded the @onsol/tldparser to the latest version which now returns the owner of the nft if the domain is wrapped as an nft.

Another small update from ANS protocol, we built a bridge to bring (ENS) .eth domains to Solana through Wormhole x ANS protocol.

those are NFTs from the get go. that means people could send stuff to .eth on solana 😆.

### DIFF
--- a/packages/app-mobile/src/components/NFTCard.tsx
+++ b/packages/app-mobile/src/components/NFTCard.tsx
@@ -12,14 +12,16 @@ export function BaseCard({
   imageUrl,
   subtitle,
 }: {
-  onPress: () => void;
+  onPress: (name?: string) => void;
   imageUrl: string | null;
   subtitle?: { name: string; length: string };
 }): JSX.Element {
   const theme = useTheme();
   return (
     <Pressable
-      onPress={onPress}
+      onPress={() => {
+        onPress(subtitle?.name);
+      }}
       style={{
         width: "50%",
         padding: 4,
@@ -42,9 +44,9 @@ export function BaseCard({
           style={{
             backgroundColor: theme.custom.colors.nav,
             position: "absolute",
-            bottom: 8,
-            left: 8,
-            right: 8,
+            bottom: 4,
+            left: 4,
+            right: 4,
             borderRadius: 8,
             paddingHorizontal: 8,
             paddingVertical: 6,
@@ -59,9 +61,11 @@ export function BaseCard({
           >
             <Text
               numberOfLines={1}
+              ellipsizeMode="tail"
               style={{
                 justifyContent: "space-between",
                 color: theme.custom.colors.fontColor,
+                maxWidth: 100,
               }}
             >
               {subtitle.name}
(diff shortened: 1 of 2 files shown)


peterpme:
hi there @crypt0miester, i temporarily removed tldparser (https://github.com/coral-xyz/backpack/pull/3365) bc it doesn't work on mobile via expo

we can move this discussion over to your repo. in the mean time, here's what we need:
- [getHashedName](https://github.com/onsol-labs/tld-parser/blob/main/src/utils.ts#L67) should be an async function.

this change should have no impact on consumers of the library. [in the 2 places that its called](https://github.com/search?q=repo%3Aonsol-labs%2Ftld-parser%20getHashedName&type=code), there are already async functions included

wrapping it in node's utils.promisify should work great!

- include a [separate index.native.js ](https://github.com/LinusU/react-native-get-random-values) (a diff example linked so you can see how) file that can polyfill `crypto.createHash` using `expo-crypto`. expo-crypto will work with both react-native and crypto

the following code should be functionally equivalent:

```ts
const encryptedActionSecret1 = crypto
      .createHash("sha256")
      .update(bytes, "utf-8")
      .digest("utf-8") as bytes;

    const encryptedActionSecret2 = (await Crypto.digestStringAsync(
      Crypto.CryptoDigestAlgorithm.SHA256,
      Buffer.from(bytes, "utf-8").toString()
    )) as bytes;
```

if you have any questions please lmk. thank you!

crypt0miester:
> hi there @crypt0miester, thank you for this!
> 
> I temporarily removed tldparser (#3365) bc it doesn't work on mobile via expo
> 
> we can move this discussion over to your repo. in the mean time, here's what we need:
> 
> 1. [getHashedName](https://github.com/onsol-labs/tld-parser/blob/main/src/utils.ts#L67) should be an async function.
> 
> this change should have no impact on consumers of the library. [in the 2 places that its called](https://github.com/search?q=repo%3Aonsol-labs%2Ftld-parser%20getHashedName&type=code), there are already async functions included
> 
> wrapping it in node's utils.promisify should work great!
> 
> 2. include a [separate index.native.js ](https://github.com/LinusU/react-native-get-random-values) (a diff example linked so you can see how) file that can polyfill `crypto.createHash` using `expo-crypto`. expo-crypto will work with both react-native and crypto
> 
> a separate index.native.js file will be consumed by expo/react-native automatically without having an impact on your web users
> 
> the following code should be functionally equivalent:
> 
> ```ts
> const encryptedActionSecret1 = crypto
>       .createHash("sha256")
>       .update(bytes, "utf-8")
>       .digest("utf-8") as bytes;
> 
>     const encryptedActionSecret2 = (await Crypto.digestStringAsync(
>       Crypto.CryptoDigestAlgorithm.SHA256,
>       Buffer.from(bytes, "utf-8").toString()
>     )) as bytes;
> ```
> 
> if you have any questions please lmk. thank you!

awesome. thanks for the quick reply. I'm on it.

crypt0miester:
hello @peterpme,

I updated [getHashedName](https://github.com/onsol-labs/tld-parser/blob/8c43d8144c653012d6a6280f234077921b5e9921/src/utils.ts#L66) and asyncified it and awaited everywhere. 

used what bonfida did by using @ethersproject/sha2 because it is easier since ethersproject already includes the shims. 

is this what you're looking for? 

please update me. because I will need to push to npmjs if it is. 

thank you btw.

Issue #3366, opened 2023-03-15
avatar loading is slow


armaniferrante:
https://user-images.githubusercontent.com/6990215/225445573-93b4a7bd-5a38-43fc-a59a-bf210816a592.mov

armaniferrante:
Note: this is only happening on my account right now because I have a lot of friends and messages.

wentokay:
can we remove the timestamp cachebuster from them?

each time we forcibly break the cache like that we're doing a DB request and making cloudflare images fetch, re-resize and reprocess the source image

hkirat:
Potentially fixed by https://github.com/coral-xyz/backpack/pull/3371 . We were loading a lot of `LocalImageManagers` all trying to cache images causing the thread to be used by the canvas API.

armaniferrante:
Appears fixed!

PR #3382, opened 2023-03-16
Fix image loads

hkirat:
PR asks for small images when small image is needed
PR modifies the `skeleton loader` logic a little to fallback to original image after 1.5s. This is done because the `onLoad` event doesn't seem to fire sometimes on the image causing the loader to stay up indefinitely


hkirat:
Related https://github.com/coral-xyz/backpack/issues/3366

PR #3388, opened 2023-03-16
add eslint rule to enforce operationName in zeus queries

wentokay:
- warns if there is no `{ operationName: string }` as a second parameter in zeus calls
- `eslint --fix` will automatically create it and using the name of the parent function as a default value
- warns if the `operationName` is not `[a-zA-Z]`

https://user-images.githubusercontent.com/101902546/226071315-67c566ec-ec74-4ed8-a5b3-8f1cebc91cef.mov


packages/eslint-config-custom/index.js
```diff
@@ -16,8 +16,10 @@ module.exports = {
     "@typescript-eslint",
     "simple-import-sort",
     "mui-unused-classes",
+    "zeus-custom",
   ],
   rules: {
+    "zeus-custom/require-operation-name": "warn",
```
wentokay, line 22: set to `warn` by default for now but will switch to `error` for autofixing in the near future

PR #3389, opened 2023-03-17
add auth.public_keys.is_primary computed field

wentokay:
Screenshot 2023-03-16 at 18 31 06

adds an `is_primary: boolean` computed field to `public_keys`, it effectively does the same as `user_active_publickey_mappings` but in a more optimized way

### results from an example large query in prod

before (~10s query) | after (~50ms query)
---|---
Screenshot 2023-03-16 at 21 04 35|Screenshot 2023-03-16 at 21 05 48


wentokay:
merging this in so it doesn't fall behind with migrations but i'll add a follow up PR to actually use the `public_keys.is_primary` fields

PR #3422, opened 2023-03-20
Expose useNavigateToChat

hkirat:
PR exposes the hook useNavigateToChat which takes two inputs - 
1. collection id - collection id of the chat to join
2. nft id - mint of the nft that gives the user access to the collection


ph101pp:
Hmm this doesnt help my case yet since its still on the frontend (although easier to use now), 
I need the logic to be in the `background` so I can make an RPC call to it from an xnft 

```
await background.request({
      method: UI_RPC_METHOD_NAVIGATION_OPEN_CHAT,
      params: [chatId, mintAddress],
    });
    ```

Issue #3211, opened 2023-03-07
white dev backpack icon not visible on default windows chrome theme


wentokay:
<img width="256" alt="Screenshot 2023-03-07 at 10 32 00" src="https://user-images.githubusercontent.com/101902546/223517068-ced6bd22-36bf-4766-bd01-109c2d73ce49.png">

on hover
<img width="182" alt="Screenshot 2023-03-07 at 10 31 56" src="https://user-images.githubusercontent.com/101902546/223517069-f1ef2850-3256-4cc4-970d-aca1e310c6c1.png">

armaniferrante:
p3 as it's not user facing. If anyone is facing this though, feel free to fix it immediatley!

asknishant:
@armaniferrante do we need to replace white color with some other color?

shivam2562001:
I have fixed this issue. can u assign me this issue? I am adding my PR here. please have a look merge it.

Shivansh-243:
is this closed because it is still not resolved 
please assign this to me 
![Screenshot 2024-02-07 211005](https://github.com/coral-xyz/backpack/assets/125989942/c73cf4c4-2fd0-4002-a318-dec824c5f5da)
here after hover
![Screenshot 2024-02-07 211054](https://github.com/coral-xyz/backpack/assets/125989942/d4177c22-848e-480e-b1fa-d4bb78635656)

PR #3468, opened 2023-03-29
icon of app-extension is fixed

shivam2562001:
I have fixed  icon issue in chrome extensions.

Issue Tag: white dev backpack icon not visible on default windows chrome theme #3211


wentokay:
hey @shivam2562001 this won't work unfortunately as [webpack will copy](https://github.com/coral-xyz/backpack/blob/64b3d08ed07519f403b0cf63e1a475b1a8222d81/packages/app-extension/webpack.config.js#L266-L271) `anchor-production.png` or `anchor-development.png` to `anchor.png` depending on whether or not it is in the production (yarn build) or development (yarn start) environment.

We currently use a red and white icon to differentiate the production and development build icons respectively.

What we really need here is to have an icon that will still be visible on the default white browser toolbar in windows. I think what we might want is black outline or perhaps a different colored icon so that it's still visible.

@trucy @aarongillett do you have any preference for the development build icon color related to #3211? I thought adding an outline might be useful in case a users' browser is the exact same color as the icon, right now we're assuming that nobody is using a bright red browser theme for instance.

aarongillett:
Maybe just add a grey or red outline for now (if you haven’t already). We can revisit with more consideration and proper cross-browser testing later.

yarn.lock
```diff
@@ -19987,10 +19987,10 @@ __metadata:
 
 "eslint-plugin-zeus-custom@file:./eslint-plugin-zeus-custom::locator=eslint-config-custom%40workspace%3Apackages%2Feslint-config-custom":
   version: 0.0.1
-  resolution: "eslint-plugin-zeus-custom@file:./eslint-plugin-zeus-custom#./eslint-plugin-zeus-custom::hash=a463b5&locator=eslint-config-custom%40workspace%3Apackages%2Feslint-config-custom"
```
hkirat, line 19990: Can u revert this?

shivam2562001: reverted

shivam2562001:
I have added black color icon for anchor-development


anchor-development

shivam2562001:
image

trucy:
@shivam2562001 here's a version of the logo with red outline and white background. 

@johnrees I can make any color versions you might need (black outline or red outline transparent fill etc) just lmk!

backpack logo (white with red outline)

shivam2562001:
yeah red outline looks cool I will add this icon in manifest

shivam2562001:
I have added above outline icon. I think this fits with the requirement.

ss

can someone pls approve this pr

aarongillett (commented):
This icon is horizontally stretched. Let’s use the icon Trucy provided [here](https://github.com/coral-xyz/backpack/pull/3468#issuecomment-1494782731).

shivam2562001:
I used the same icon with previous dimension of 72x72

Let me know should I use the same dimension as trucy provided?

PR #3495, opened 2023-03-30
Added URLfor activating hover effect

bonysureliya:
The Anagram logo was not having hover effect ,
Then I read the code,
So there was this if statement that said if URL of the image is empty so don't hover So I have added URL so you can upto the and check if the URL is correct or not

### DIFF
--- a/web/components/Partners.tsx
+++ b/web/components/Partners.tsx
@@ -25,7 +25,7 @@ const data = [
   },
   {
     image: '/brands/anagram.svg',
-    url: ''
+    url: 'https://www.anagram.care/'
   }
 ];
 


web/components/Partners.tsx
```diff
@@ -25,7 +25,7 @@ const data = [
   },
   {
     image: '/brands/anagram.svg',
-    url: ''
+    url: 'https://www.anagram.care/'
```
armaniferrante, line 28: This is the incorrect url. In fact, I don't think the anagram referenced here has a site.

bonysureliya: So what if we add condition on navigation link rather 
Example :
``` javascript 
if (url =="") {
  // Add hovereffect but not the url
}
 ````

This would help to remove inconsistency

bonysureliya: if you find this approach useful , then I will start working on adding this block of code

hkirat:
Closing for now since low priority

PR #3501, opened 2023-03-30
update tldparser version to version 0.5.3

tonivlad:
Hi @peterpme 

We made the changes you requested here ([#3379](https://github.com/coral-xyz/backpack/pull/3379#issuecomment-1472174412))

* functions are all async and TldParser uses ethersproject crypto functions instead of crypto 
* we built our own expo app and the new version of TldParser works using the same shims that you have in packages/app-mobile. [app-mobile/index.js](https://github.com/coral-xyz/backpack/blob/master/packages/app-mobile/index.js) + [app-mobile/src/crypto-shim.ts](https://github.com/coral-xyz/backpack/blob/master/packages/app-mobile/src/crypto-shim.ts)

this is the exact shims that we used:

```js
// Inject node globals into React Native global scope.
global.Buffer = require('buffer').Buffer;
global.process = require('process');

if (typeof c === 'undefined') {
  global.btoa = function (str) {
    return new Buffer(str, 'binary').toString('base64');
  };
}

if (typeof atob === 'undefined') {
  global.atob = function (b64Encoded) {
    return new Buffer(b64Encoded, 'base64').toString('binary');
  };
}

if (__DEV__ && typeof global.crypto !== 'object') {
    global.crypto = {
        getRandomValues: (array) => array.map(() => Math.floor(Math.random() * 256)),
    };
}
```

* added the function to parse ANS domains in useIsValidAddress.ts

thank you for for all your help @peterpme and @armaniferrante 

and WAO!


peterpme:
Hey @tonivlad,

I really appreciate you taking the time to try and make this work!

That being said, the Backpack app is crashing upon testing this out. Here's what I did

Reproduction steps:
- installed `@onsol/tldparser@0.5.1` in `packages/app-mobile/package.json`
- Call `useIsValidAddress` with `monkey.sol`

Result:
- The function itself responds with values, but claims monkey.sol is false
- The UI of the app doesn't load.

We'd really love to merge this in so I've created a simple screen you can test this on in our app (see below). If you run into any issues please let me know.

Thanks again!

```tsx
import { View, Text } from "react-native";

import { Blockchain } from "@coral-xyz/common";
import {
  useAnchorContext,
  useEthereumCtx,
  useIsValidAddress,
} from "@coral-xyz/recoil";

export function TestingScreen() {
  const { provider: solanaProvider } = useAnchorContext();
  const ethereumCtx = useEthereumCtx();

  const data = useIsValidAddress(
    Blockchain.SOLANA,
    "monkey.sol",
    solanaProvider.connection,
    ethereumCtx.provider
  );

  return (
    <View
      style={{
        flex: 1,
        backgroundColor: "yellow",
        alignItems: "center",
        justifyContent: "center",
      }}
    >
      <Text>isValidAddress: {JSON.stringify(data, null, 2)}</Text>
    </View>
  );
}
```

tonivlad:
Hey @peterpme 

updated tldparser to version 0.5.3

can you please test this again and tell us if the problem still exists?

crypt0miester:
@peterpme,

attached is the branch for the tests that I am runnning.  
https://github.com/onsol-labs/backpack/tree/onsol-test/packages/onsol-test

even if I remove TldParser, from useIsValidAddress() it will never work. because the recoil atomValues does not get set from some reason. I have no idea why it returns empty values.

which means that the error is not in the tld-parser library rather in something else when not using app-mobile (i.e. idk maybe, the recoil states not being set due to the wallets not being initiated/registered)

regards,
miester.

Issue #3476, opened 2023-03-29
Whitespace causes invalid private key


trucy:
If user adds a space when they paste and continue when importing a wallet from private key, they get an error saying it’s invalid

armaniferrante:
we can just strip the whitespace easy enough

Soul-Eater69:
Hi trucy, can I take this up?

armaniferrante:
Assigned.

PR #3504, opened 2023-03-30
Fixed whitespace causes invalid private key 

Soul-Eater69:
Closes https://github.com/coral-xyz/backpack/issues/3476

This happens when the user enters a space and then pastes the private key, the fix I've added is to trim off the spaces from the input while setting the state variable.


armaniferrante:
Thank you! Please remove the yarn.lock changes and we can merge this!

Soul-Eater69:
Thanks, Armani. I have removed the yarn changes and committed the same.

PR #3546, opened 2023-04-02
Make Sure that the two users in a DM Chat don't have the same color

### LINKED ISSUE #3459: DMs should always have different username colors
Usernames are currently randomly assigned, but sometimes you can end up with the same color for sender and recipient in a DM. Should be a rule (only for DMs) that makes sure the two colors are unique for sender and receiver.

Example of them being the same

Screenshot 2023-03-28 at 12 42 28 PM

RashulChutani:
This PR makes sure that users in a DM Chat don't have the same color.
Fixes #3459 
It works by checking if the `uniqueUserIds` has a length of 2, which
implies that the chat is between 2 users. It further checks if the users
have been assigned the same `colorIndex`. If it is the case, it changes it
for one of them.

### DIFF
--- a/packages/tamagui-core/src/hooks/chat/useChatsWithMetadata.tsx
+++ b/packages/tamagui-core/src/hooks/chat/useChatsWithMetadata.tsx
@@ -3,6 +3,7 @@ import type {
   SubscriptionType,
   UserMetadata,
 } from "@coral-xyz/common";
+import { NEW_COLORS } from "@coral-xyz/common";
 import { useChats, useUser } from "@coral-xyz/recoil";
 
 import { useUsersMetadata } from "./useUsersMetadata";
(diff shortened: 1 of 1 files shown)


RashulChutani:
Hi @getusha Thanks for suggesting the refactors. I have pushed the suggested changes. Thanks.

packages/tamagui-core/src/hooks/chat/useChatsWithMetadata.tsx
```diff
@@ -30,10 +31,24 @@ export const useChatsWithMetadata = ({
...
+  if (uniqueUserIds.length === 2) {
+    const [userId1, userId2] = uniqueUserIds;
 
+    function haveSameColor(user1, user2) {
+        return (user1 && user2) && user1.colorIndex === user2.colorIndex;
+    }
+
+    if (haveSameColor(users[userId1], users[userId2])) {
+        const newColorIndex = (users[userId1].colorIndex + 1) % NEW_COLORS.length;
+        users[userId2].colorIndex = newColorIndex;
```
hkirat, line 48: You're updating the return value of a react hook here. As a general rule you should never update the value returned from a hook/selector

hkirat (changes requested):

RashulChutani:
I also tested my changes, and it seems to be working fine. :smile: In the below Image, the `colorIndex` for `User2` has been updated to be different from `User1` (It's new value is `1 + User1.colorIndex`) . 
diff_colors
Thanks.

packages/tamagui-core/src/hooks/chat/useChatsWithMetadata.tsx
```diff
@@ -34,20 +35,35 @@ export const useChatsWithMetadata = ({
...
   console.log("debug2", { chats, userIds, uniqueUserIds });
 
+  // Make Sure that both the users in a DM chat do not have the same color
+  const colorIndexMap = new Map();
+  if (uniqueUserIds.length === 2) {
+    const [userId1, userId2] = uniqueUserIds;
+
+    if (haveSameColor(users[userId1], users[userId2])) {
+      const newColorIndex = (users[userId1].colorIndex + 1) % NEW_COLORS.length;
+      colorIndexMap[userId2] = newColorIndex;
```
hkirat, line 45: you should probably use colorIndexMap.set(userId2, newColorIndex) here. 
Similarly use .get below when you're fetching it. 
Rest all looks good.

RashulChutani:
Hi @hkirat, I have added the changes requested. Thanks.

hkirat (approved):

hkirat:
Thankyou @RashulChutani for your contribution!

RashulChutani:
> Thankyou @RashulChutani for your contribution!

My Pleasure! 🙂

PR #3554, opened 2023-04-03
Now Derivation path is displayed under wallet details(Only when Devel…

### LINKED ISSUE #3360: developer mode: show derivation path in wallet details
In developer mode only. Can add it as a section here.

Screen Shot 2023-03-14 at 6 26 37 PM

mrb1nary:
Derivation path is now shown to the user under wallet details page only when the Developer Mode is enabled.
Fixed #3360


armaniferrante:
This needs significant work. Closing to avoid clutter. Please re-open and please test before opening a PR.

PR #3585, opened 2023-04-04
Clear backpack.#publicKey on disconnect

### LINKED ISSUE #3555: Not an approved origin error when signing after reconnecting
Repro with the wallet-adapter example app: https://solana-labs.github.io/wallet-adapter/example/

- Use any connect button to connect backpack
- Click eg sign message, verify it works correctly
- Disconnect then connect to backpack again
- Use any of the sign message/transaction buttons, you get an error:

```
VM368 react_devtools_backend.js:2655 WalletSignMessageError: https://solana-labs.github.io is not an approved origin
    at Je.He [as signMessage] (_app-cb484cb16f6d065c.js:74:470325)
    at async _app-cb484cb16f6d065c.js:74:476273 Je
```

mcintyre94:
This fixes a bug where the connect function is not called properly on a reconnect, as #publicKey is already defined

Specifically this is because of this condition in the wallet-standard wrapper (https://github.com/coral-xyz/backpack/blob/master/packages/wallet-standard/src/wallet.ts#L149):
```
    #connect: ConnectMethod = async ({ silent } = {}) => {
        if (!silent && !this.#backpack.publicKey) {
            await this.#backpack.connect();
        }

        this.#connected();

        return { accounts: this.accounts };
    };
```

On the first connect, `this.#backpack.publicKey` is undefined and `this.#backpack.connect()` is called correctly

However, on the second connect `this.#backpack.publicKey` was already set and so `this.#backpack.connect()` was not called. `this.#connected()` was called, and since this uses `this.#backpack.publicKey` it was able to emit an event that made it look to the wallet adapter UI like a connection was successfully performed. But the internal code in `this.#backpack.connect()` hadn't been called, and so signing didn't actually work.

The fix is to set `backpack.#publicKey` to undefined on disconnect, so that `this.#backpack.connect()` is called correctly on the reconnect.

Closes #3555

### DIFF
--- a/packages/provider-core/src/provider-solana.ts
+++ b/packages/provider-core/src/provider-solana.ts
@@ -161,6 +161,7 @@ export class ProviderSolanaInjection
       params: [],
     });
     this.#connection = this.defaultConnection();
+    this.#publicKey = undefined;
   }
 
   async openXnft(xnftAddress: string | PublicKey) {


armaniferrante:
Thank you!

Issue #3594, opened 2023-04-04
Update link color in notifications for friend requests


trucy:
<img width="376" alt="Screenshot 2023-04-04 at 12 56 44 PM" src="https://user-images.githubusercontent.com/321395/229878273-5be4f4ad-a77d-4fa7-804a-e7c0e0ec3e5d.png">

Currently `blue` in light mode and `rgb(52, 152, 219)` for dark mode, these should be:
- #0057EB or rgb(0,87,235) for light mode
- #4C94FF or rgb(76,148,255) for dark mode

Anubhav311:
Hi trucy, I would like to do this one. Can you please assign it to me.

Merdi-kim:
@trucy is that about the "Accept" button?

trucy:
@Merdi-kim yes! will edit the poorly worded issue name :)

spantheslayer:
Hey @trucy , Created a PR for this issue https://github.com/coral-xyz/backpack/pull/3597

This PR updated the values of the color for the `Accept` link color

hsnice16:
Hi @trucy 
I have raised a PR that will resolve this issue without updating UI at any other place, as I did not changed the colour on global level. 
Kindly review and merge that PR.
Thank you!

Anubhav311:
Hi @trucy 
the color of "Accept" on this screen should also be change?
<img width="487" alt="Screenshot 2023-04-06 at 10 44 08 PM" src="https://user-images.githubusercontent.com/30749439/230449376-c16802d3-a747-4a62-8f98-1831d478180f.png">

coderdeadpool:
Hey @trucy, Is this issue open to everyone? 
Can i work on this issue ?

Anubhav311:
@trucy got it... just pushing the change

Anubhav311:
I've updated the blue color globally. Please take a look: https://github.com/coral-xyz/backpack/pull/3602

manancodes:
Some PRs created for this issue were closed without giving any reason. Should I start working on this?

Chandradeep-Pra:
Hi Trucy , I would want to take this issue , can you please assign it to me ?

Mindslayer001:
Hey @trucy, Is this issue open to everyone?
Can i work on this issue ?

Manish-Raz:
Is the issue still open ?

Aditya2829Pal-B:
i think this issue no more open

PR #3602, opened 2023-04-04
updated color of link for accepting friend request

Anubhav311:
This PR updates the color of the `Accept` link button for friend request notifications. 

A new `acceptFriendRequest` color is created as a global variable for reusability and convention. 

I have not changed the `blue` color as it is being used by some other components also.

https://github.com/coral-xyz/backpack/issues/3594


Anubhav311:
As per the updated info, I've changed the global variable `blue` instead of creating a new variable.
Can you please have a look @callensm

packages/themes/src/colors.ts
```diff
@@ -53,7 +53,7 @@ export const DARK_RED_BORDER_MED = "rgba(241,50,54,0.4)";
 
 // NOTE: Do not include anything but colors in here. No box shadows, borders, etc.
 export const DARK_COLORS: CustomColors = {
-  blue: "#3498db",
+  blue: "#4C94FF",
```
hkirat, line 56: Hi can you confirm this isn't being used other places?
Since this'll update everywhere `theme.custom.colors.blue` is being used

packages/themes/src/colors.ts
```diff
@@ -124,7 +124,7 @@ export const DARK_COLORS: CustomColors = {
 
 // NOTE: Do not include anything but colors in here. No box shadows, borders, etc.
 export const LIGHT_COLORS: CustomColors = {
-  blue: "blue",
+  blue: "#0057EB",
```
hkirat, line 127: same comment as above

hkirat (changes requested):

PR #3622, opened 2023-04-05
Fix truncate wallet name

Merdi-kim:
As addressed in issue #2462 , this pull request solves the problem by implementing a method that truncates names longer than 10 characters. This number was chosen by me because wallet names of less than 10 characters did not break the UI.


Merdi-kim:
@getusha there is a way of doing that with CSS but for a given long name like _this is my name_  the end result would look more like _this is m..._ instead of the end goal that is desired which is _this is m...e_

PR #3641, opened 2023-04-07
return exact username match as first result when searching for users

wentokay:
fixes #3641

before | after
-------|-----
Screenshot 2023-04-07 at 15 21 36|Screenshot 2023-04-07 at 15 29 26

Screenshot 2023-04-07 at 15 38 33


wentokay:
merged in #3646

Issue #2571, opened 2023-02-08
disable features requiring a JWT when not authenticated


tomlinton:
Features requiring JWT should display an empty state saying they are unavailable when the user is not authenticated.

Blocked by https://github.com/coral-xyz/backpack/issues/2570

tomlinton:
`useAuthenticatedUser` is available and will be null if there is no JWT

wentokay:
merging into #2617

PR #3689, opened 2023-04-13
enable as much functionality as possible if any backend infra is down

### LINKED ISSUE #2617: Ensure backends going down doesn't lock the wallet
Right now if `backend` goes down we dont let the user unlock the wallet.
They should be allowed to use the wallet as usual

wentokay:
fixes #2617

WIP

### TODO

- [x] allow password log in
- [x] allow user switching
- [x] load balances page
- [x] send and receive tokens
- [x] load collectibles page
- [x] disable added features in collectibles page
- [x] send and receive collectibles
- [x] load xNFTs
- [ ] use `useAuthenticatedUser` where needed https://github.com/coral-xyz/backpack/issues/2571
- [ ] ensure invalid authentication is still handled correctly
- [ ] handle all errors gracefully
- [ ] decide what to do with xNFTs that use JWT functionality
- [ ] generate QR code locally?
- [ ] notifications?
- [ ] swaps?
- [ ] show disconnected message in chat?
- [ ] show that user is not authenticated somehow
- [ ] tracking referrals/swap rewards


kevinfaveri:
hey @wentokay need help here? can you get me up to speed on the real current status and what needs to be done?

Issue #3692, opened 2023-04-13
Wallet address is still shortened when there's enough space to display the full address


wentokay:
to reproduce - 

click `Receive` button on the balances screen
popout backpack extension (ctrl+g)
expand window or make it full screen

<img width="920" alt="Screenshot 2023-04-13 at 19 51 17" src="https://user-images.githubusercontent.com/101902546/231855995-13d93b1e-081b-4303-bf29-fc228dbbe29e.png">

adarshlilha:
would like to take this up!

wentokay:
assigned you @adarshlilha, thanks

Anubhav311:
Hi @wentokay 
Sorry I was already working on this one as it wasn't assigned to anybody.
I've opened a PR. Can you please take a look.

PR #3694, opened 2023-04-13
Added responsiveness to wallet address truncation

Anubhav311:
Using CSS for responsive string truncation for wallet address. Earlier It was hard coded using javascript.

One difference to be noted is that the ellipsis (...) is now placed at the end of the wallet address. Earlier it was placed in the middle of the wallet address.

Before:
Screenshot 2023-04-14 at 4 50 58 AM

After:

https://user-images.githubusercontent.com/30749439/231902939-3987ece1-a076-4ee4-96c5-8fd1abe17abb.mov

Closes https://github.com/coral-xyz/backpack/issues/3692


wentokay:
Hey @Anubhav311 thanks for your PR but unfortunately we must show the final characters of an address for a few reasons.

One of the biggest ones is that as it's a global standard of sorts to truncate at the middle. Users need to be able to see both the initial and final characters when verifying addresses if they want to transfer between other other software or hardware wallets.

This approach will useful for dynamic wallet name truncation and stuff like that though so the effort is still appreciated :)

I looked into a couple of options -

- this might be promising, https://codepen.io/markchitty/pen/RNZbRE although it might be a challenging to replicate the behavior in react native (we're building a [mobile app](https://github.com/coral-xyz/backpack/tree/master/packages/app-mobile) in parallel with the extension, and being able to share code between platforms is a valuable time-saver for us)
- [`text-overflow-middle: ellipsis`](https://www.w3.org/wiki/Text-overflow_middle_cropping) would have been ideal although it seems like it never got adopted as a standard

Anubhav311:
Thank you @wentokay for sharing the relevance of middle truncation for wallet address. I didn't know about it.

The way it is done in the codepen that you shared, I think it will be hard to do it with the `<TextField />` in our case. We might have to replace it with some divs and spans.

I'll give it a try.

Anubhav311:
Hi @wentokay can you please take a look and let me know if this is good. Or should I make any changes. It's truncating at the middle of the string.

https://user-images.githubusercontent.com/30749439/232117291-1c346480-a4fa-4b8c-9766-fe51c3751f16.mov

PR #3719, opened 2023-04-18
Update to use tensor.trade over Magic Eden when No NFT's are detected

cloakd:
ME won't handle the load for the MADLads mint. lets promote a web3 friendly markeplace


aarongillett:
We’ll be moving to a new design that shows multiple marketplace options.

cloakd:
Cool, when? Seems like a good stopgap in the mean time.

Issue #3729, opened 2023-04-19
Message button in chat screen doesn't seems to do anything if its just a message to be displayed this shouldn't be a button.


MohitMandalia:
### Message button in chat screen doesn't seems to do anything if its just a message to be displayed this shouldn't be a button.

Could I get some design inputs here to make correction if this is not intentional.

<img width="369" alt="Screenshot 2023-04-19 at 7 32 56 PM" src="https://user-images.githubusercontent.com/36087806/233100396-52f67420-c6e6-47a4-9a39-0a5d1651a692.png">

Happy to take this up.

aarongillett:
Agreed; this shouldn’t be a button. Thanks for spotting this. Let’s remove the button fill and just keep the icon as a non-interactive illustration, like so:

<img width="487" alt="image" src="https://user-images.githubusercontent.com/4157473/233166959-706fc736-7977-4ea3-970b-a5c04b5c7039.png">

PR #3736, opened 2023-04-19
Make the message button in the chat screen an icon rather than a button.

MohitMandalia:
### What
- Change the message button on chat screen when chat is empty to a normal icon.

### Why
- As the button doesn't do anything it makes sense to make it just an icon, so that user is not confused.

### How
- Remove the IconButton and change the Icon color to match below  design

### Closes
-  #3729


hkirat:
Hi , the background that `IconButton` bring is nice. 
Could you either make it unclickable (make the cursor normal) or add a background to the icon?

For ref this is what it looks like
Screenshot 2023-04-26 at 6 52 20 PM

MohitMandalia:
HI @hkirat I basically suggested 2 things:
- Make background not clickable.
- Remove the background.

in the issue #3729 @aarongillett suggested me to remove the background and gave a design to follow, hence I have made the color as per @aarongillett 's suggestion.

I could change and add background and make it not clickable, let me know what is the correct thing you guys want to follow.
Thanks

Issue #1521, opened 2022-11-17
Add try catches around all `toString` methods


hkirat:
Seems like a common theme of the app breaking is `Cant read property toString of undefiend` .
We should wrap all `toString` calls in a try catch so that the app doesn't break, and if it does we know where exactly the issue lies

More details - https://discord.com/channels/985994296337498182/1035923763570675832/1042574330808242278

tomlinton:
It seems like we should be able to fix this by working on our typing a bit, particularly removing not null assertions.

hkirat:
Yeah better types is definitely the cleaner way to do this

SlokAks:
I would like to work on this issue. Please assign it to me. Thanks.

hkirat:
@SlokAks go for it!

DeepakSR-12:
Hi @hkirat . If the assignee is not working on this, can I please take it over?

SlokAks:
Hi @DeepakSR-12, was busy with office work for last week. I am currently working on it.

PR #3758, opened 2023-04-22
wrap all toString calls in try catch

SlokAks:
Issue #1521 
Please review the changes, @hkirat. This should fix the issue.

Implemented a wrapper, safeToString() to wrap toString() call in a try-catch block.


armaniferrante:
As discusssed in the original issue, we should do this with better typing instead of try catch. Closing. Thank you for the PR!

Issue #3776, opened 2023-04-24
Update Collectibles empty state to include multiple marketplaces


aarongillett:
### UI
![image](https://user-images.githubusercontent.com/4157473/234124879-4627e271-b515-4694-acca-8e565d085d0c.png)

### Icon
Interests Rounded
https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Ainterests%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048

### Assets
![tensor](https://user-images.githubusercontent.com/4157473/234128675-93bdc1f3-600d-49dd-bd88-a9d835983387.svg)
![magic_eden](https://user-images.githubusercontent.com/4157473/234127390-0e9f3789-987c-4e33-a45a-5c77ea0da398.svg)
![hyperspace](https://user-images.githubusercontent.com/4157473/234127129-dc0433d0-2517-4db3-80cc-7b182d968930.svg)
![blur](https://user-images.githubusercontent.com/4157473/234127231-eef518c2-65de-4ace-ae49-ab5efa810ead.jpg)
![opensea](https://user-images.githubusercontent.com/4157473/234127234-81da6e04-9b6d-4123-a327-bb020577e9b8.svg)

### Links
https://www.tensor.trade
https://magiceden.io
https://hyperspace.xyz
https://blur.io
https://opensea.io

Merdi-kim:
@aarongillett I would like to work on this issue, can I?

aarongillett:
Sure thing, thanks @Merdi-kim !

ritvij14:
@aarongillett can I work on this issue? If yes, just to confirm, on the empty screen I will have to add this so that people can go to those platforms for collectibles, correct?

PR #3783, opened 2023-04-25
Update collectibles empty state

Merdi-kim:
Here are the results after working on issue #3776. 

Among the provided images, one was a **.jpeg** format so I had to turn it into an SVG file to standardize all the marketplaces icons 
Screenshot 2023-04-25 at 14 49 33
Screenshot 2023-04-25 at 14 49 42


aarongillett:
This looks good, thanks @Merdi-kim. Two small points:
- can we use the rounded version of the material icon?
- can we make the material icon have a thinner 2px border? This looks like ~4px.

Merdi-kim:
Do you mean the icon above **No collectibles** ??

aarongillett:
> Do you mean the icon above **No collectibles** ??
> 
> Concerning the weight of that icon, I do not think there is a CSS property that can reduce it. Maybe if I can look for a more light icon it could solve. Or simply if I can have access to the same icon in the design that could help a lot @aarongillett

Yes, that’s the icon I meant. You can use the rounded version and adjust the weight of the border using the properties defined [here](https://fonts.google.com/icons?selected=Material%20Symbols%20Rounded%3Ainterests%3AFILL%400%3Bwght%40200%3BGRAD%400%3Bopsz%4048). Thanks @Merdi-kim !

Merdi-kim:
@aarongillett  I have made the change to the icon and here is how it looks now 
Screenshot 2023-04-27 at 11 29 39

aarongillett:
Thanks @Merdi-kim . The icon looks better, but still a little dark compared to the mockup. Let’s go with something lighter, like `LIGHT_BORDER_COLOR = "#DFE0E6"` and its dark equivalent (via [colors.ts](https://github.com/coral-xyz/backpack/blob/master/packages/themes/src/colors.ts))

Merdi-kim:
@aarongillett I have just updated the changes you requested me to do. Now here is the result of how the icon looks like 


Screenshot 2023-04-28 at 13 47 34

PR #3791, opened 2023-04-25
Fixes friend request notification not updating

### LINKED ISSUE #3710: Messages Requests: accepting friends via the chat does not clear the notifications for new friend requests. I find myself approving friend requests twice.
image

atulya2109:
# Description 

The friend request notifications weren't updating when the friend request was accepted through the chat. This was primarily because when sending or accepting a request through the chat window, only `requested` was being updated in the recoil state.

## Proposed Fix
Check if `friendshipValue.remoteRequested` is set to true for the `ChatScreen`, in that case, update state to set `areFriends` to true. If `remoteRequested` is false, then update `requested` to true.

https://user-images.githubusercontent.com/20705464/234343554-2ef92cba-679e-46f2-8915-41615f9dbf62.mov

Fixes #3710

### DIFF
--- a/packages/message-sdk/src/components/ChatScreen.tsx
+++ b/packages/message-sdk/src/components/ChatScreen.tsx
@@ -1,4 +1,5 @@
 import { ChatRoom } from "@coral-xyz/chat-sdk";
+import { updateFriendshipIfExists } from "@coral-xyz/db";
 import { useFriendship, useUpdateFriendships } from "@coral-xyz/recoil";
 
 export const ChatScreen = ({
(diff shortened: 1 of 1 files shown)


packages/message-sdk/src/components/ChatScreen.tsx
```diff
@@ -27,9 +28,9 @@ export const ChatScreen = ({
       }}
     >
       <ChatRoom
-        type={"individual"}
+        type="individual"
         remoteUsername={username}
-        username={""}
+        username=""
```
hkirat, line 33: Can you run the lint locally once? This seems incorrect the way it is formatted

hkirat (changes requested):

hkirat (approved):

Issue #3778, opened 2023-04-25
[BUG] Images presenting fully stretched as max height/width is not set when using backpack wallet as a standalone Chrome window


kevinfaveri:
<img width="1594" alt="Screenshot 2023-04-25 at 02 51 55" src="https://user-images.githubusercontent.com/18222497/234148534-e31063ac-f201-43f5-b764-083419cf6c16.png">

kevinfaveri:
Attempts at improving backpack desktop wallet NFT vieweing experience here: https://github.com/coral-xyz/backpack/pull/3792

Closing this issue, we can decide on the PR to do it or not.

PR #3792, opened 2023-04-25
fix: streched image dimensions

kevinfaveri:
### Summary
This PR attempts at improving, mainly, the desktop experience of using Backpack, that is currently like this: https://github.com/coral-xyz/backpack/issues/3778

It stretches small images as width: 100% is always enforced

This PR changes width: 100% to maxWidth: 100%; and center the item (which is useful if the NFT image is small, like SMB). Mobile version remains unaffected by the changes as it is a really small window, so continue to prob just use 100%.

### Implementation
https://www.loom.com/share/0364a5513b44446caad0e7c085f43050 (non-popup mode didnt record lol, i guess i need to record entire desktop for it to show).


armaniferrante:
Thanks Kevin!

PR #3795, opened 2023-04-25
using alpine image , using multi stage build

AtulRajput01:
In this  dockerfile using multi-stage builds, we first use a smaller base image FROM node:18-alpine instead of FROM node:18.15.0@sha256:eee4c74962a855bc1694676cd5069460e545a8580e0e574b2c59eae4f40bd5bd. This reduces the size of the image and improves its security & also reduce the number of layer in docker development file.

In the first stage, we copy the required files and install the dependencies using Yarn. Then, we build the required components of the application 

In the second stage, we copy the required files from the first stage into a new base image, node:18-alpine, and set the working directory to the backend-ws directory.


AtulRajput01:
> Hi @AtulRajput01 , thankyou for your contributions Are you able to compile the image locally? I'm getting the following error while doing the same Screenshot 2023-04-26 at 6 20 41 PM

@hkirat yes sir i am able to compile the image locally.
Screenshot from 2023-05-05 11-44-14           

i think you are getting the  error while doing encountered an error while building Docker image for the project  in the path backpack/docker/backend-ws.
However, successfully created a Dockerfile for the same project on the path coral and able to run the Docker image on local machine.
Possible issues with Docker configuration in the backpack/docker/backend-ws path.

Issue #3812, opened 2023-04-27
Use higher-quality avatar image on lock screen


aarongillett:
The lock screen avatar is low quality (120x120px) and pixelated on retina.

Suggestion: use a `@2x` asset for HQ screens.

Related: [Cache avatar on lock screen #3733](https://github.com/coral-xyz/backpack/issues/3733)

<img width="375" alt="image" src="https://user-images.githubusercontent.com/4157473/234831691-d2c50656-6545-4127-8b98-b2d890395b80.png">

Abhijit20112003:
@wentokay I just create a PR

wentokay:
now fixed, longer term changes including [faster loading](https://github.com/coral-xyz/backpack/issues/3733) coming soon

PR #3813, opened 2023-04-27
Use higher-quality avatar image on lock screen #3812

Abhijit20112003:


wentokay:
closing as this would double the on-screen size of the pfp and the fix for #3812 is now live, thanks for your contribution though

Issue #3739, opened 2023-04-20
Broken search result view


MohitMandalia:
## Issue
- The search result view breaks when the search text is too long.

## Suggestions
- Should we only allow 15 char as the username won't be long enough or add `...` if search exceeds certain char limit what do you think?
  
<img width="378" alt="Screenshot 2023-04-20 at 6 29 23 AM" src="https://user-images.githubusercontent.com/36087806/233230937-fd2915bd-8a57-440f-a144-b67a976bc5f3.png">

Happy to work on this.

aarongillett:
Thanks for spotting this. I’d suggest we add an ellipsis `…` to truncate the result (this is what Google does).

psugandhraj123:
Hyy @MohitMandalia. Are you still working on this. If not I can take this up.

MohitMandalia:
Yup I am already on it.

PR #3832, opened 2023-04-28
fix: search text in friends screen

MohitMandalia:
#### What
- As search result view breaks when the search text is too long, add properties to the text such that it has ellipsis `...` at the end if the search text is long.

#### Why
-  To make the UX better for the user and void breaking the views.

#### How
- Add ellipsis when the search text is too long.

#### MR changes:

Screenshot 2023-04-28 at 6 35 30 AM

#### Closes
- #3739


MohitMandalia:
callouts: if the text has whitespace it will allow the user to search for longer text and can got down till the end of the view.

Do we want to do any changes to this??

Issue #3821, opened 2023-04-27
Unable to reply to most recent message


aarongillett:
Hovering over the most recent message and trying to reply is buggy:

https://user-images.githubusercontent.com/4157473/234967896-e223ed67-d7ce-49dc-9610-ad5b83234f5e.mov

Merdi-kim:
Can I fix this @aarongillett?

aarongillett:
Closing this as it’s been fixed by #3848

PR #3848, opened 2023-04-29
Fixed bug which prevents user to reply to most recent message

Kritik-J:
issue: #3821 

As we are using opacity instead of visibility on Jump to bottom button, So it overlays on the most recent message and prevents to reply

Before: 

https://user-images.githubusercontent.com/82360525/235285030-0ca80fa7-f800-48a4-afa8-6274ef75b9c0.mp4

After: 

https://user-images.githubusercontent.com/82360525/235285044-a3d2124c-551e-4a30-aed3-755eb105cd2d.mp4


armaniferrante:
Thank you!

PR #3976, opened 2023-05-16
feat: support token-2022 accounts, transfers, and burns

joncinque:
#### Problem

Token-2022 has been out for some time, but backpack doesn't support it.

#### Solution

Add basic support for Token-2022! This PR boils down to a few things:

* fetch accounts for token-2022
* store the program id on accounts and mints after they're fetched
* use the program id for ATA derivation, transfers, and burns

Since the `Program` type from Anchor doesn't allow modifying the program id, I've gone with a `TokenInterface` intermediate type, which contains a `withProgramId` function, to expose a `Program<SplToken>` with the correct program id. This may be a little heavy, so let me know if you prefer something else. You could also modify instructions as they're created, but going through the `TokenInterface` makes it hard to get wrong, which I wanted to prioritize.

Note that this PR doesn't touch wrapped SOL at all. There is a wrapped SOL mint on token-2022, but this keeps all wrapped SOL on the old token program for now.

Also, I noticed a bug with burning NFTs. My test account wasn't an ATA and failed because the burn function assumed that the source was an ATA, so I added the `source` parameter in the burn request. If you prefer to have that work done separately, let me know.

Edit: I've also written a little guide at https://github.com/solana-labs/solana-program-library/pull/4325 if you need more info

### DIFF
--- a/packages/app-extension/src/components/Unlocked/Nfts/NftDetail.tsx
+++ b/packages/app-extension/src/components/Unlocked/Nfts/NftDetail.tsx
@@ -598,6 +598,7 @@ function BurnConfirmationCard({
       const _signature = await Solana.burnAndCloseNft(solanaCtx, {
         solDestination: solanaCtx.walletPublicKey,
         mint: new PublicKey(nft.mint.toString()),
+        programId: new PublicKey(nft.programId.toString()),
         amount,
       });
       setSignature(_signature);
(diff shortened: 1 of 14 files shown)


callensm:
@joncinque have all of the new additions for this PR been through e2e testing?

joncinque:
> @joncinque have all of the new additions for this PR been through e2e testing?

I tested showing / transferring fungible tokens, and showing / burning / transferring a basic NFT. I didn't test the Cardinal / Programmable / OpenCreator NFTs, but since the base case worked with the new parameter, the hope is that they'd all receive it properly, even if the protocol doesn't support token-2022 yet.

Zahra73mh:
9U2QJLPw9W2WnsqUgXg5h5LQ3uRNB1SWH1SHCxoVZw7t

PR #3979, opened 2023-05-17
fix checking versioned transaction status

### LINKED ISSUE #3978: xNFTs do not support checking the status of a versioned transaction
After submitting a versioned transaction and waiting for the confirmation, an error will be thrown and the `window.xnft.solana.sendAndConfirm` promise fails

Screenshot 2023-05-17 at 10 23 56

wentokay:
fixes #3978 

`window.xnft.solana.sendAndConfirm` fails at the confirmation stage because versioned transactions aren't supported in the getParsedTransaction config, this adds support by adding `maxSupportedTransactionVersion: 0`

### DIFF
--- a/packages/background/src/backend/solana-connection.ts
+++ b/packages/background/src/backend/solana-connection.ts
@@ -50,13 +50,14 @@ import type {
   GetProgramAccountsConfig,
   GetProgramAccountsFilter,
   GetSupplyConfig,
+  GetTransactionConfig,
+  GetVersionedTransactionConfig,
   InflationGovernor,
   InflationReward,
   LeaderSchedule,
   LogsCallback,
   LogsFilter,
   Message,
-  MessageV0,
   NonceAccount,
   ParsedAccountData,
   ParsedConfirmedTransaction,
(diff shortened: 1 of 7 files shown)


packages/common/src/solana/background-connection.ts
```diff
@@ -1010,17 +1013,23 @@ export async function confirmTransaction(
...
+    let tx = await c.getParsedTransaction(txSig, {
+      commitment,
+      maxSupportedTransactionVersion: 0,
+    });
     while (tx === null) {
-      tx = await c.getParsedTransaction(txSig, commitment);
+      tx = await c.getParsedTransaction(txSig, {
+        commitment,
+        maxSupportedTransactionVersion: 0,
+      });
```
wentokay, line 1032: and this is the other part

packages/background/src/backend/solana-connection.ts
```diff
@@ -472,7 +455,8 @@ export class SolanaConnectionBackend {
     );
     return {
       context: {
-        slot: tx.slot,
+        // tx! keeps existing behavior, but it should be tx?
+        slot: tx!.slot,
```
wentokay, line 459: and this change

wentokay, line 459: and this change

armaniferrante:
Can we get this in?

wentokay:
I'll do some more manual testing to try and ensure it doesn't introduce regressions, but if you could also quickly glance over it before merging it'd be much appreciated 🙇‍♂️

wentokay:
converted back to draft as this needs a bit more work to support swaps and possibly other transactions

wentokay:
Screenshot 2023-05-31 at 19 36 16

more context, the connection needs the confirmed commitment to be specified, the confirmation commitment preferences screen might cause issues

Screenshot 2023-05-31 at 19 37 41

PR #4075, opened 2023-05-29
Eliminate race condition in `useIsAddressValid`

steveluscher:
Eliminate race condition in `useIsAddressValid`
## Summary:

In general, async code in React effects is unsafe. To be precise, _setting state_ after an awaitable is unsafe unless you track whether the effect that triggered you is still ‘relevant’ (ie. its dependencies haven't changed _and_ the component in which it lives has not unmounted).

```ts
useEffect(() => {
  (async () => {
    const result = await somethingThatTakesTime(foo);
    setResult(result); // `foo` might have changed by now!
  })();
}, [foo]);

// 1. `foo` gets set to `bar`
// 2. `somethingThatTakesTime` gets called for `bar`
// 3. `foo` changes to `baz`
// 4. `somethingThatTakesTime` gets called for `baz`
// 5. `somethingThatTakesTime` returns a result for `baz` and `setResult` gets called
// 6. `somethingThatTakesTime` returns a result for `bar` and `setResult` gets called
//
// Now `foo` is `baz`, but you're displaying the _result_ for `bar`. UI: corrupt.
```

In this PR we implement the tracking necessary to prevent this; we cancel in progress work whenever `address` changes. This will prevent a situation where the result from a _previous_ address ends up overwriting the result for the _current_ address.

## Test Plan:

Years of watching this bite developers in the ass.

### DIFF
--- a/packages/eslint-config-custom/shared/react.js
+++ b/packages/eslint-config-custom/shared/react.js
@@ -60,7 +60,7 @@ module.exports = {
     "react-hooks/exhaustive-deps": [
       "warn",
       {
-        additionalHooks: "(useRecoilCallback|useRecoilTransaction_UNSTABLE)",
+        additionalHooks: "(useAsyncEffect|useRecoilCallback|useRecoilTransaction_UNSTABLE)",
       },
     ],
   },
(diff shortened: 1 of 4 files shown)


steveluscher:
Two things to note:

1. This is probably a problem in a _lot_ more places than just this. I didn't check.
2. A way more elegant solution is to use https://github.com/rauldeheer/use-async-effect, but I hand rolled this PR presuming that you wouldn't want to take on another dependency.

PR #4102, opened 2023-06-01
token-2022: Pass in program id when deriving ATA

joncinque:
#### Problem

Token-2022's great, but ATAs are even better, and the first pass wasn't passing in the program id when deriving the ATA, causing problems when creating someone else's ATA.

#### Solution

Pass in the program id too!

### DIFF
--- a/packages/common/src/solana/index.ts
+++ b/packages/common/src/solana/index.ts
@@ -194,7 +194,8 @@ export class Solana {
           walletPublicKey,
           destinationAta,
           destination,
-          mint
+          mint,
+          programId
         )
       );
     }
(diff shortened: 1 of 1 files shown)


packages/common/src/solana/index.ts
```diff
@@ -632,7 +634,8 @@ export const generateWrapSolTx = async (
         walletPublicKey,
         destinationAta,
         destination,
-        NATIVE_MINT
+        NATIVE_MINT,
+        TOKEN_PROGRAM_ID
```
joncinque, line 638: This one's pedantic, but will hopefully make things clearer if you decide to support wrapped SOL on token-2022

PR #4127, opened 2023-06-05
service worker fixes with secure-background etc

peterpme:
- wip
- wip tbd


packages/app-mobile/src/App.tsx
```diff
@@ -37,110 +36,88 @@ export function App(): JSX.Element {
 
   return (
     <ErrorBoundary fallbackRender={renderError}>
-      <Suspense fallback={<FullScreenLoading />}>
+      <BackgroundHiddenWebView />
```
peterpme, line 39: for anybody tuning in, the order here matters

PR #4162, opened 2023-06-16
tweak mobile nav bar titles

wentokay:


packages/app-mobile/src/navigation/AccountSettingsNavigator.tsx
```diff
@@ -153,11 +153,11 @@ export function AccountSettingsNavigator(): JSX.Element {
           component={YourAccountScreen}
           options={{
             title: "Your Account",
-            headerBackTitle: "Profile",
```
wentokay, line 156: @peterpme is it ok to remove this for the moment? It's showing "Profile" in the back section when every other page shows "Settings"

PR #4187, opened 2023-06-20
fix mobile wallets settings button

wentokay:
fixes this error that was thrown when trying to navigate to a wallet's settings as blockchain wasn't included as the first parameter

Screenshot 2023-06-20 at 11 54 21

### DIFF
--- a/packages/app-mobile/src/screens/Unlocked/EditWalletsScreen.tsx
+++ b/packages/app-mobile/src/screens/Unlocked/EditWalletsScreen.tsx
@@ -1,4 +1,3 @@
-import type { Blockchain } from "@coral-xyz/common";
 import type { Wallet } from "@coral-xyz/recoil";
 
 import { Suspense, useCallback } from "react";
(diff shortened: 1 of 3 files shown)


packages/app-mobile/src/screens/Unlocked/EditWalletsScreen.tsx
```diff
@@ -84,19 +87,18 @@ function WalletList2({ onPressItem }) {
...
+      onPressItem={(blockchain, { name, publicKey, type }) => {
+        navigation.navigate("edit-wallets-wallet-detail", {
+          blockchain,
+          publicKey,
+          name,
+          type,
+        });
+      }}
+    />
+  );
```
wentokay, line 101: I made this function inline so it didn't need to be retyped, I assume there isn't a performance penalty here as it's only rendering one WalletList2 and was already a reference to a nested function. Lmk if you want to avoid this approach though @peterpme

peterpme: love the idea, thanks!! sorry about the regression

wentokay:
closing in favor of #4192

PR #4267, opened 2024-01-02
Implement downgrading non-changing writable accounts to readonly during transaction signing

Jac0xb:
### PR

- Related https://twitter.com/armaniferrante/status/1742079119448494248

- Not interested in a job, nor have I ever looked or contributed to this repo so this implementation could be jank but I think wallet draining as become a public health crisis that needs to be stopped.

- Implement downgrading writable accounts to read if no change in account state is found.

- Not sure if toly/Armani had something else in mind for comparing the changes in account state just grabs account info for all writable accounts and compares to post-simulation account states.

- I think there are flaws to this approach: largest being that an attacker simply partially signs the transaction before sending to wallet thus rendering it unchangeable. (which pink drainer kit already does).

- Another issue is if a transaction fails to simulate the simulation accounts return `null` but the transaction could still succeed, what should be done in this indeterminate state.

- Was unable to find logic that checks if a transaction is modifiable but assumed its checked before config (priority fee / compute units) are added.

- Haven't tested yet, will figure that out tomorrow, if someone wants to finish this PR they are welcome to.


packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -423,6 +431,75 @@ export function useSolanaTxData(
...
+    ) {
+      const determineDowngradeWritableAccounts = async () => {
+        setLoading(true);
+
+        const writableAccounts = simulationAccounts.map(
+          (k) => new PublicKey(k)
+        );
+        const preSimulationWritableAccounts =
+          await connection.getMultipleAccountsInfo(writableAccounts);
+        const simulation = await solanaClient.wallet.simulate({
```
Jac0xb, line 449: Not quite sure what the difference between connection / solanaClient is but saw solanaClient being used in the commented code below which called this simulate function so figured thats what I should use.

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -423,6 +431,75 @@ export function useSolanaTxData(
...
+      simulationAccounts.length > 0
+    ) {
+      const determineDowngradeWritableAccounts = async () => {
+        setLoading(true);
+
+        const writableAccounts = simulationAccounts.map(
+          (k) => new PublicKey(k)
+        );
+        const preSimulationWritableAccounts =
+          await connection.getMultipleAccountsInfo(writableAccounts);
```
Jac0xb, line 448: I dont forsee this ever being bigger than the 100 account rpc limit but its technically possible with lookup tables.

packages/secure-background/src/services/svm/util.ts
```diff
@@ -80,6 +81,16 @@ export async function sanitizeTransactionWithFeeConfig(
     }
 
     transaction.add(...coreIxs);
+
+    // Not sure where the logic that checks if the transaction is partially signed but
```
Jac0xb, line 85: I made an assumption about this being checked if its a  modifiable transaction since its being modified with compute unit / priority fee instructions.

packages/recoil/src/hooks/useTransactionData.tsx
Jac0xb: Need to figure out if I should use connection.simulateTransaction or solanaClient.wallet.simulate. solanaClient.wallet.simulate is messy because it adds the signer to the front of the simulation accounts making the ordering of the simulationAccounts misaligned with preSimulationWritableAccounts

Jac0xb:
Relying on account info rpc call to compare with simulation account is prone to race conditions where state changes outside of the transaction are taken as state changes by the account change logic.

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -423,6 +431,75 @@ export function useSolanaTxData(
...
+      solanaClient &&
+      transaction &&
+      walletPublicKey &&
+      simulationAccounts.length > 0
+    ) {
+      const determineDowngradeWritableAccounts = async () => {
+        setLoading(true);
+
+        const writableAccounts = simulationAccounts.map(
+          (k) => new PublicKey(k)
```
saberistic, line 445: are all simulation accounts writable accounts?

packages/recoil/src/hooks/useTransactionData.tsx
```diff
@@ -423,6 +431,75 @@ export function useSolanaTxData(
...
+              pubkey: writableAccounts[i],
+              account: preSimulationWritableAccounts[i],
+            };
+          })
+          .filter(({ pubkey, account: presimulationAccount }, i) => {
+            if (!presimulationAccount) {
+              return false;
+            }
+            const simulationAccount = simulation.accounts![i];
+            if (simulationAccount === null && presimulationAccount == null) {
```
saberistic, line 479: since we checked truthiness of preSimulationAccount in L475 will this ever be true?

saberistic:
You are fast @Jac0xb, find this issue pretty interesting but I don't have much experience with this repo either. Forgive in advance if comments are out of context trying to wrap my head around it

stegaBOB:
Another potential big issue with this is the writable constraint in Anchor. Last I checked it does actually validate the account is writable, so if that constraint wasn't needed 100% of the time this would probably fail. IMO that constraint check should be removed. Can't think of a case where it's actually useful to check at runtime other than for IDL suggestions.

stegaBOB:
> I checked it does actually validate the account is writable, so if that constraint wasn't needed 100% of the time this would probably fail.

Confirming that this is indeed the case. https://github.com/coral-xyz/anchor/blob/47ff77f4df503b109e32a559daa209ade5f86487/lang/syn/src/codegen/accounts/constraints.rs#L245

ph101pp:
Thank you for this contribution @Jac0xb ! I cherry picked your commit into our internal repo and added UI to allow disabling the feature during signing. Assuming we don't run into unexpected issues during testing, this should be part of the next release 🎉

PR #2335, opened 2023-01-30
Added Resolver for Solana and Ethereum Domain Name Websites[REVISED]

mjkid221:
# Solana and Ethereum Name Service Resolver for Backpack.
<hr/>
EDIT:
<b>
This is a revised version of the pull request at #2259. I ran into some issues with the older build in #2259 so I started a fresh PR.
<b/>
<hr/>

Solana and Ethereum Domain Name Services (DNS) provide a decentralized and cost-effective way to link domain names to corresponding on-chain data, such as websites associated with those domain names.

This pull request adds support for resolving Solana (SNS) and Ethereum (ENS) domain names in the URL, allowing users to automatically redirect to their website stored in the domain content. The feature build add support for IPFS, IPNS, and regular HTTPS websites (and even IP addesses!), making Backpack a more versatile and desirable crypto wallet. This implementation was greatly inspired from `zhvng`'s [sns-resolver](https://github.com/zhvng/sns-resolver) which was deprecated. 

The new feature integrates with the existing wallet functionality, and utilizes the Solana (Bonfida) and Ethereum (Ethers) library to lookup domain name content and redirect to attached websites upon validation. All invalid URLs will redirect to an error 404 page. Users can also configure the gateway to use of their own choice (including custom url) to resolve their domain website if desired so.

This is probably the first time someone has ever created a domain name website resolver for Solana at the wallet level, and I believe it will greatly enhance the user experience and make Backpack a more powerful tool for the crypto community.

Please review the code and let me know if there are any issues or suggestions for improvements. I would be happy to make any necessary changes to ensure this feature gets merged into the master branch. Thank you!

## Process to Use
1. Set content for SNS or ENS with URL of your desired website. 
    - You can resolve to either IPFS OR IPNS via setting the field to either one of `ipfs=`, `ipns=`, `ipfs://`, `ipns://` respectively. 
    - You can also resolve to a regular `https://` or `http://` website by just setting the field with a regular url if the underlying domain name service supports it. 

2. Assuming Backpack is already installed, go to your URL tab and search for your `<domainName>.sol` or `<domainName>.eth`.
   - It will automatically redirect as well if the search domain is prefixed with `http://` or `https://` such as `http://<domainName>.eth` or postfixed with `/` such as `<domainName>.sol/`.
   - Redirects also work with both SOL and ETH subdomains but it is important to note that:
       -  ENS allows for multiple nested child subdomains which is resolvable by the implementation. Technically, it should be able to resolve infinite(not really) amount of child subdomains. 
[...]

PR #4284, opened 2024-02-17
Web DNS Resolver for SNS + ENS V2

mjkid221:
# Purpose
Revived an old feature from https://github.com/coral-xyz/backpack/pull/2335 which adds Web domain name resolutions via SNS and ENS.

Understandable if the team wants to bench this while they work on other top priority items. Just wanted to get this out of the way.

### DIFF
--- a/packages/app-extension/src/background/index.ts
+++ b/packages/app-extension/src/background/index.ts
@@ -1,5 +1,44 @@
 import { start } from "@coral-xyz/background";
 
+import { supportedDomains, urlPatterns } from "../dns-redirects/constants";
+import { redirect } from "../dns-redirects/helpers/tabHelper";
+
 start({
   isMobile: false,
 });
+
+/**
+ * Resolves domain names in the form of URLs.
+ */
+chrome.webNavigation.onBeforeNavigate.addListener(
+  async (details) => {
+    await redirect(details.url);
+  },
+  {
+    url: supportedDomains.map((domain) => {
+      return { urlMatches: `^[^:]+://[^/]+.${domain}/.*$` };
+    }),
+  }
+);
+
+/**
+ * Resolves domain names in the form of browser searches via Google, Bing, etc.
+ * DuckDuckGo has a unique search pattern and must be queried separately.
+ */
+chrome.webNavigation.onBeforeNavigate.addListener(
+  async (details) => {
+    const domainUrl = new URL(details.url).searchParams.get("q");
+    if (domainUrl && domainUrl.indexOf(" ") < 0) await redirect(domainUrl);
+  },
+  {
+    url: supportedDomains.flatMap((param) =>
+      urlPatterns.map((pattern) => {
+        return {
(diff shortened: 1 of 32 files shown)


packages/background/src/backend/core.ts
```diff
@@ -642,6 +642,61 @@ export class Backend {
     return SUCCESS_RESPONSE;
   }
 
+  async domainContentIPFSGatewayRead(uuid: string): Promise<string> {
+    const data = await secureStore.getWalletDataForUser(uuid);
+    return data.webDnsResolutionGateway.ipfsGateway;
+  }
+
+  async domainContentIPFSGatewayUpdate(ipfsGateway: string): Promise<string> {
+    const uuid = (await this.keyringStore.activeUserKeyring()).uuid;
```
adasarpan404, line 651: `const userUUID = await this.keyringStore.activeUserKeyring()?.uuid;`

It will great if you use **Nullish Coalescing Operator**(?) here.

adasarpan404: Not neccessary just saying.

packages/app-extension/src/background/index.ts
```diff
@@ -1,5 +1,44 @@
 import { start } from "@coral-xyz/background";
 
+import { supportedDomains, urlPatterns } from "../dns-redirects/constants";
+import { redirect } from "../dns-redirects/helpers/tabHelper";
+
 start({
   isMobile: false,
 });
+
```
adasarpan404, line 9:
```
import { start } from "@coral-xyz/background";
import { supportedDomains, urlPatterns } from "../dns-redirects/constants";
import { redirect } from "../dns-redirects/helpers/tabHelper";

// Start the background process
start({
  isMobile: false,
});

// Event listener for handling URL navigation
chrome.webNavigation.onBeforeNavigate.addListener(async (details) => {
  try {
    // Extract the domain URL
    const url = new URL(details.url);
    const domainUrl = url.searchParams.get("q") || url.hostname;

    // Check if the domain URL contains spaces
    if (domainUrl && !domainUrl.includes(" ")) {
      // Redirect to the resolved URL
      await redirect(domainUrl);
    }
  } catch (error) {
    console.error("Error occurred during navigation:", error);
  }
}, {
  // Define URL patterns to match
  url: supportedDomains.flatMap((param) =>
    urlPatterns.map((pattern) => ({
      urlMatches: pattern.includes("duckduckgo")
        ? `${pattern}\\.${param}$`
        : `${pattern}\\.${param}&.*$`,
    }))
  ),
});
```
Instead of having two separate chrome.webNavigation.onBeforeNavigate event listeners, we can combine them into one to reduce redundancy.

adasarpan404: just a suggestion, please verify on your own.

adasarpan404: I have put an suggestion for the change in whole file for this

PR #4289, opened 2024-02-22
Support Vietnamese language

trquangtien:
**### Purpose**
Backpack now has huge support from the Vietnamese community => Please consider supporting the Vietnamese language.
Adding Vietnamese language translation for the locale i18n file.


git2024abc:
Can Japanese be supported?

armaniferrante:
Please note we are in the middle of redoing our entire i18n system. So these changes won't apply. We will absolutely support both Vietnamese and Japanese. Thank you for the PR in any case!

Issue #3202, opened 2023-03-07
don't allow wallet names to start with @


wentokay:
as it could cause confusion with usernames in future

noobie7:
Can this be assigned to me?

wentokay:
it's yours @noobie7! thanks

shubhiscoding:
hii @wentokay @armaniferrante  please review this PR #4306 
In this PR I have made necessary  changes to avoid user from naming there wallet starting from '@'
![image](https://github.com/coral-xyz/backpack/assets/95687419/52a75035-4413-44e8-a20d-9b989fcac3bd)

shubhiscoding:
#4306

Ananmayd:
hey can I work on this issue ?

DhruvWebDev:
Hey can i work on this issue...
see I am able to rename my wallet, starting with @walletname

![Image](https://github.com/user-attachments/assets/5f3817fe-549c-4177-b9dd-1708b3605120)

bharat-dussa:
If this issue still persist Can i work on this issue ?

PR #4306, opened 2024-04-03
don't allow wallet names to start with @

shubhiscoding:
solves #3202
wouldn't allow users to keep wallet names that start with @
image


shubhiscoding:
solves #3202 , please review @wentokay

PR #4310, opened 2024-05-03
setState change to correct manner. update useStep.tsx

karthik11135:
Changed the way setState is used. When the state is dependent upon on the val, function format must be used.

### DIFF
--- a/packages/app-extension/src/hooks/useSteps.tsx
+++ b/packages/app-extension/src/hooks/useSteps.tsx
@@ -2,9 +2,9 @@ import { useState } from "react";
 
 export const useSteps = () => {
   const [step, setStep] = useState(0);
-  const nextStep = () => setStep(step + 1);
+  const nextStep = () => setStep((prevStep) => prevStep + 1);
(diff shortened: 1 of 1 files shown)


packages/app-extension/src/hooks/useSteps.tsx
```diff
@@ -2,9 +2,9 @@ import { useState } from "react";
 
 export const useSteps = () => {
   const [step, setStep] = useState(0);
-  const nextStep = () => setStep(step + 1);
+  const nextStep = () => setStep((prevStep) => prevStep + 1);
   const prevStep = () => {
-    if (step > 0) setStep(step - 1);
+    if (step > 0) setStep((prevStep) => prevStep - 1);
   };
```
steveluscher, line 8:
```tsx
  const prevStep = () => setStep((prevStep) => prevStep > 0 ? prevStep - 1 : prevStep);
```

React will not re-render if the state transition results in the same value before and after.

steveluscher (approved):

PR #4344, opened 2024-09-19
typescript error fixed for Image component

### LINKED ISSUE #4342: Just a Typescript bug [not too serious]
So in the backpack/web/components/AppNoLink.tsx 
image
This Image says that Type '"100px"' is not assignable to type 'SafeNumber' 

1. I think we should remove the *px from there that would be a ez fix.

Keyur3766:
fixed #4342

- Removed "px" from the height and width of the images to prevent typescript error.

### DIFF
--- a/web/components/AppNoLink.tsx
+++ b/web/components/AppNoLink.tsx
@@ -12,8 +12,8 @@ function AppNoLink({ iconUrl, name }: AppNoLinkPros) {
             blurDataURL="/brands/aurory.jpg" //TODO: fix me
             placeholder="blur"
             quality={50}
-            width="100px"
-            height="100px"
+            width="100"
+            height="100"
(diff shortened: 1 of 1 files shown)


web/components/AppNoLink.tsx
```diff
@@ -12,8 +12,8 @@ function AppNoLink({ iconUrl, name }: AppNoLinkPros) {
             blurDataURL="/brands/aurory.jpg" //TODO: fix me
             placeholder="blur"
             quality={50}
-            width="100px"
-            height="100px"
+            width="100"
```
akashchouhan16, line 15: shouldn't it be `100vw` and `100vh` instead?

akashchouhan16: my bad, it doesn't take anything else part from `Number`

